PR #110, opened 2026-04-24
Audio: exclude commentary/descriptive audio from DV lighter-track swap

rrgomes:
## Summary

On Dolby Vision content, `preferredLighterAudioTrack` swaps a heavy-decode
primary track (TrueHD / DTS-HD) for a lighter codec (E-AC-3 / AC-3 / AAC) so
the read loop can keep up with 4K DV throughput. The previous filter matched
only on language, so it could silently pick a commentary or audio-description
track as the "main" audio.

This PR filters out alternate-purpose tracks (commentary, audio description,
narration, sign language, karaoke, isolated score) at selection time, prefers
the Plex `isDefault`-flagged candidate among what remains, and returns `nil`
if no appropriate lighter track exists — so the pipeline keeps the heavy
track rather than playing wrong audio confidently.

## Reproduction

Any UHD Blu-ray rip with both a default English TrueHD Atmos 7.1 track and an
English AC-3 2.0 commentary / audio-description track. Pre-fix behavior:
commentary plays as if it were the main mix.

## Test plan

- Built and installed on Apple TV 4K (2nd gen), tvOS 26.4.
- Verified correct default audio on a known-affected title (DV UHD with
  commentary track) — was previously playing commentary, now plays the main
  Atmos mix.
- Verified non-DV content unaffected (unchanged code path).
- Verified titles without commentary/description tracks unaffected.

---

_Prepared with LLM assistance and verified on-device._

### DIFF
--- a/Rivulet/Services/Plex/Playback/Pipeline/DirectPlayPipeline.swift
+++ b/Rivulet/Services/Plex/Playback/Pipeline/DirectPlayPipeline.swift
@@ -2461,11 +2461,18 @@ final class DirectPlayPipeline {
         return heavy.contains(where: { normalized == $0 || normalized.hasPrefix($0) })
     }
 
-    /// Find a lighter audio track for DV throughput. Prefers same language,
-    /// prioritizes EAC3 > AC3 > AAC by channel count.
+    /// Find a lighter audio track as a substitute for a heavy-decode primary
+    /// on DV content. Excludes special-purpose tracks (commentary, audio
+    /// description, narration); prefers same language; strongly prefers
+    /// `isDefault`-flagged tracks; tie-breaks by channel count then
+    /// EAC3 > AC3 > AAC. Returns nil when no suitable lighter track exists —
+    /// in that case the caller must keep the heavy track rather than swap to
+    /// a commentary or audio-description track as if it were the main audio.
     private func preferredLighterAudioTrack(than current: FFmpegTrackInfo) -> FFmpegTrackInfo? {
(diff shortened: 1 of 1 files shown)


l984-451:
@claude please review this PR

l984-451:
@rrgomes thanks for these PRs!

Issue 1 flagged above - is the intention to prefer stereo with the +100?

rrgomes:
@l984-451 yes, intentional — the +100 is meant to dominate channel count when both signals are present, on the read that the Plex-flagged default is the user's deliberate signal for "main audio" while channel count is a heuristic.

In practice the asymmetric case Claude flagged (stereo default + 7.1 non-default) is rare in a typical Plex library: most multi-track items either don't set `isDefault` at all (in which case the bonus is moot and channel count rules) or set it on the same track that's already the highest-channel one (in which case +100 just reinforces what `channels * 10` was already doing). The combinations where +100 actually matters and the channel-count asymmetry is large are mostly libraries where someone has gone out of their way to mark a specific stereo track as the main audio — which seems like the right thing to honor.

That said, if you'd rather the channel-count gap survive a default-flag for big spreads (e.g. you've seen real libraries with stereo-default + 7.1-non-default that should resolve to 7.1), happy to drop the bonus to +50 on a follow-up. The numeric weight is a knob; the semantic ranking — "default flag wins ties" — is the part I want to keep.

l984-451 (approved):
Conservative, well-targeted fix. Verified on-device per the test plan. LGTM.

PR #112, opened 2026-04-27
Player: prefer Plex extendedDisplayTitle in in-playback track menu

rrgomes:
## Summary

Plex tags audio and subtitle streams with both a short `displayTitle`
("English (AC3 5.1)") and a long-form `extendedDisplayTitle`
("English (AC3 5.1) - Director's Commentary"). The extended form is
what distinguishes a commentary or audio-description track from the
main mix; without it, multi-track items show two rows whose
label-side text is identical, and the user has to guess.

The in-playback (swipe-down) menu rendered tracks fully synthesized
from language + codec — never even displaying `displayTitle`, let
alone the extended form. `PlexStream` already parses
`extendedDisplayTitle` and the `VideoInfoOverlay` already prints it,
so the data is end-to-end available — it just wasn't propagated
through `MediaTrack`.

## Changes

- Add `extendedDisplayTitle: String?` to `MediaTrack`; populate from
  `PlexStream` and at the FFmpeg-track enrichment site in
  `UniversalPlayerViewModel`.
- In `PlayerControlsOverlay`, prefer `extendedDisplayTitle` for the
  title row of audio + subtitle entries; fall back to the existing
  synthesized `audioFormatString` / `languageDisplay` form when Plex
  doesn't supply one (typical for embedded streams without titles, or
  FFmpeg-only tracks before enrichment).
- Subtitle row (synthesized language / codec form) preserved beneath
  the descriptive title — Plex Web shows both.

## Test plan

- [x] Multi-track UHD HEVC HDR DV item with TrueHD + AC-3 commentary
      and multi-track subs (Doctor Strange in the Multiverse of Madness)
      — commentary row now distinguishable from main mix
- [x] Item with no `extendedDisplayTitle` — falls back to prior
      synthesized form
- [x] FFmpeg-only tracks (pre-enrichment) — still use synthesized form
- [x] Build + run on Apple TV 4K (2nd gen), tvOS 26.4

🤖 Authored with assistance from Claude (Anthropic).

PR #114, opened 2026-04-28
Detail page: pre-play audio/subtitle pickers with server-side persistence

rrgomes:
## Summary

Two changes bundled because they only become end-to-end useful
together:

1. **Pre-play track pickers on the detail page** — circular sibling
   buttons (audio + subtitles) on the movie/episode action HStack.
   Selecting a track persists server-side via Plex's
   per-user-per-part stream selection, so the choice applies to
   subsequent plays from any client (Plex Web, mobile, our app),
   mirroring first-party behavior.
2. **Selection-priority fix** — Plex's per-item explicit selection
   now wins over the language-preference managers in
   `applyAudioPreference` / `applySubtitlePreference`. Previously the
   managers returned `eng` even when nothing was stored, which let
   `findBestMatch` override deliberate per-item picks (a user who
   selected "Director's Commentary" in any client would still get
   the highest-channel English track played because TrueHD 7.1 beats
   commentary AC-3).

The two are bundled because the picker UI is mostly cosmetic without
the priority-tier fix — picks survive, but the language manager
overrides them at startup.

## Persistence model

- `PUT /library/parts/<partId>?audioStreamID=X` for audio.
- `PUT /library/parts/<partId>?subtitleStreamID=X` (with `0` for
  subtitles-off) for subs.
- Selection also stored in `@State` on the detail view as a
  same-session fast path. `presentPlayer()` awaits any in-flight
  PUT + `loadFullMetadata` refresh before constructing the player
  VM, so the HLS URL is built against fresh server-side state and
  not stale metadata.
- Plumbed via two new optional `UniversalPlayerViewModel` init
  parameters: `initialAudioTrackId: Int?` and
  `initialSubtitleSelection: InitialSubtitleSelection` — a small enum
  distinguishing "no preselection" / "explicitly off" / specific id.

## New selection priority

In `applyAudioPreference` / `applySubtitlePreference`:

1. Pre-play picker selection from this session.
2. Plex's per-item explicit selection — `selected: true` track that
   is **not** also `default: true` (and for subtitles, also not
   `forced: true`). Distinguishes deliberate user picks from
   default behavior.
3. App-level language preference (existing behavior).
4. Plex's default fallback.

## Implementation notes

- Pickers shown only on `movie` and `episode` detail pages —
  show / season / artist / album / track pages don't expose a single
  playable track list.
- Picker state resets when navigating to a different item within the
  same detail view (via collection / recommendation in-page nav).
- `TrackSelectionSheet.swift` previously existed in the repo with no
  callers; this PR is its first wiring.
- The audio path always issues `selectAudioTrackWithoutSaving` even
  when the resolved id matches `currentAudioTrackId`. Required because
  DirectPlay loads the file's default-flagged audio track on its own
  and Plex's HLS session-build doesn't always reflect the persisted
  per-part selection — an explicit switch is the only reliable way
  to honor the user's actual choice. Cost: a possibly-redundant HLS
  session rebuild at startup; correctness wins.
- Also fills `MediaTrack.extendedDisplayTitle` from `PlexStream` and
  prefers it in `MediaTrack.displayName`, so picker rows show
  "English (AC3 5.1) - Director's Commentary" instead of
  "English (AC3 5.1)" when available.

## Caveats / cross-PR notes

- **Textually overlaps #112.** Both PRs add `extendedDisplayTitle` to
  `MediaTrack` (same field, same init/`PlexStream`-init plumbing).
  Whichever merges first, the other will need a trivial rebase on
  `MediaTrack.swift` — both halves are identical so the resolution
  is just deduplication. #112 is the in-playback menu's use of the
  field; this PR additionally tweaks `displayName` to use it.
- **Subtitle persistence is end-to-end useful only with subtitle
  rendering on `main`.** Currently the `RivuletPlayer` path doesn't
  render text or bitmap subtitles on `main`; the player gets told
  the right subtitle id, but nothing appears on screen. A separate
  PR for subtitle rendering follows. Audio persistence is fully
  end-to-end on this PR alone.

## Test plan

- [x] Apple TV 4K (2nd gen), tvOS 26.4, with Doctor Strange in the
      Multiverse of Madness (UHD HEVC HDR DV + TrueHD/AC-3 commentary
      + multi-track subs)
- [x] Pick commentary audio + commentary subs on detail page → exit
      app → return to detail page → both shown as selected, `Play`
      starts in the correct tracks without re-picking
- [x] Pick a track in Plex Web → open Rivulet detail page → shows
      Plex Web's choice as selected
- [x] Subtitles-off picked → server-side persistence stores `0`
- [x] Show / season / artist pages — no picker buttons rendered

### DIFF
--- a/Rivulet/Services/Plex/Playback/MediaTrack.swift
+++ b/Rivulet/Services/Plex/Playback/MediaTrack.swift
@@ -18,6 +18,12 @@ struct MediaTrack: Identifiable, Equatable, Sendable {
     let isForced: Bool
     let isHearingImpaired: Bool
 
+    /// Plex's long-form descriptive title when present (e.g.,
+    /// "English (AC3 5.1) - Director's Commentary"). Preferred over `name`
+    /// in user-facing pickers so commentary / audio-description / SDH
+    /// tracks are distinguishable from the main mix.
+    let extendedDisplayTitle: String?
+
     // Audio-specific
     let channels: Int?
 
@@ -33,6 +39,7 @@ struct MediaTrack: Identifiable, Equatable, Sendable {
         isDefault: Bool = false,
         isForced: Bool = false,
         isHearingImpaired: Bool = false,
+        extendedDisplayTitle: String? = nil,
         channels: Int? = nil,
         subtitleKey: String? = nil
     ) {
@@ -44,6 +51,7 @@ struct MediaTrack: Identifiable, Equatable, Sendable {
         self.isDefault = isDefault
         self.isForced = isForced
         self.isHearingImpaired = isHearingImpaired
+        self.extendedDisplayTitle = extendedDisplayTitle
         self.channels = channels
         self.subtitleKey = subtitleKey
     }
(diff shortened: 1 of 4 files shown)


l984-451:
@claude please review this PR

l984-451 (approved):
Reviewed manually + applied review fixes (session usage, URLComponents, item-nav race, dead code) in 8ddbee2.

PR #115, opened 2026-04-28
Subtitles: render text + bitmap PGS/VOBSUB on RivuletPlayer path

rrgomes:
## Summary

Five-commit fix making text and bitmap subtitles actually render on
the FFmpeg / `RivuletPlayer` path (the default for direct play).
Before this PR: subtitle tracks appeared in the in-player picker, the
manager correctly collected cues from the FFmpeg read loop, but
nothing was drawn on screen. After: SRT / embedded text / PGS /
VOBSUB / DVB-SUB all render, at correct size, in the correct
position.

The four code changes are independent of one another but tightly
coupled — none alone is sufficient to render anything. A binary
refresh of the FFmpeg frameworks bundles the libavcodec rebuild.

### 1. Mount subtitle overlay on RivuletPlayer path

`SubtitleOverlayView` was gated on `viewModel.player != nil`, but
that property is the `AVPlayer` instance — nil whenever the
FFmpeg/`RivuletPlayer` path is active. The overlay never entered the
view hierarchy, so cues had nowhere to render. Fix: mount the
overlay when either player exists.

### 2. Enable bitmap subtitle decoders in the FFmpeg build

Add `pgssub`, `dvdsub`, and `dvbsub` to `--enable-decoder` so
libavcodec ships the Blu-ray PGS, DVD VOBSUB, and DVB-SUB decoders.
Without these, `avcodec_find_decoder()` returns nil for the bitmap
codec IDs and `DirectPlayPipeline`'s subtitle dispatch falls through
to the text path, where the binary subtitle data is read as UTF-8
and silently dropped.

Also fixes a latent packaging bug in `Scripts/build-ffmpeg-tvos.sh`:
`SCRIPT_DIR` was computed from `$0` after the script had `cd`'d into
the FFmpeg source tree, so relative invocations like
`./Scripts/build-ffmpeg-tvos.sh ./ffmpeg` would try to `cd` into a
non-existent path during xcframework packaging. Resolve script and
project paths once at the top of the script using
`${BASH_SOURCE[0]}`.

### 3. Include "vobsub" in DVD bitmap codec normalization

Plex reports MKV-embedded DVD bitmap subtitle tracks as codec
`"vobsub"`, while FFmpeg reports the same stream as
`"dvd_subtitle"`. The Plex/FFmpeg codec match in
`RivuletPlayer.selectEmbeddedSubtitle` runs both names through
`MediaTrack.normalizedSubtitleCodec`, but the dvdsub case only
listed `"dvdsub"` and `"dvd_subtitle"` — so `"vobsub"` fell through
to default and never matched. Add it to both display (`"VOBSUB"`)
and normalized (`"dvdsub"`) cases so all three names canonicalize
identically.

### 4. Scale bitmap rects against codec-reported reference resolution

`BitmapSubtitleRectView` previously hardcoded a 1920×1080 reference
when scaling rect coordinates onto the screen. That happens to be
correct for Blu-ray PGS, but VOBSUB rects are authored in 720×480
(NTSC) or 720×576 (PAL) DVD space, and DVB-SUB rects are typically
in 720×576 broadcast space. Treating either as 1920×1080 left
glyphs shrunk to ~37% of intended size and shifted up and left.

After `avcodec_decode_subtitle2`, `codecContext.width/height`
reflects the actual authoring resolution. Read those values,
propagate them through `DecodedSubtitleFrame` and
`BitmapSubtitleCue`, and use them at render time. Fall back to
1920×1080 if the codec reports 0 (defensive — none of the
supported codecs should leave this unset, but cheap to keep
PGS-equivalent behavior as the safe default).

## The binary refresh

The fifth commit refreshes the FFmpeg framework binaries to match
the build-config change in (2). Output of running the updated
`Scripts/build-ffmpeg-tvos.sh` against FFmpeg `n8.1` source. New
`libavcodec.a` includes the pgssub/dvdsub/dvbsub decoders enabled
in the prior commit; `libavformat`, `libavutil`, and
`libswresample` are refreshed to matching `n8.1` versions.

Also fills in the simulator slots of each xcframework. They
previously held a 520-byte ar stub (broken — simulator builds were
quietly linking against empty archives). Populated each simulator
slot with the same arm64 binary as the device slot, since the
existing build script targets arm64 only. This makes simulator
builds actually link on Apple Silicon hosts; a true x86_64
simulator slice can be added later if needed.

The diff is large (~13MB) but is mechanically reproducible from the
build-config commit by re-running `Scripts/build-ffmpeg-tvos.sh`.

## Test plan

- [x] Apple TV 4K (2nd gen), tvOS 26.4
- [x] SRT external subtitles render
- [x] Embedded text subtitles render
- [x] PGS at 1080p source — renders at correct size/position
      (previous hardcoded 1920×1080 reference happened to match)
- [x] PGS at 2160p source — codec reports 1920×1080, identical
      behavior to previous
- [x] VOBSUB — renders at correct size, matching Plex's own client
      and Infuse output (was previously ~37% size, mispositioned)
- [x] SDH ↔ non-SDH track switching works
- [x] Auto-selection from saved preferences fires on first play

## Cross-PR notes

- Pairs with #114 (pre-play track pickers): #114 persists a chosen
  subtitle id server-side, and this PR is what makes that choice
  actually visible on screen for the FFmpeg player path. #114 is
  end-to-end useful for audio alone; for subtitles, both PRs need
  to land.
- Independent of #112 / #110 / #111.


l984-451:
@claude please review this PR

l984-451 (approved):
Reviewed manually. Auto-review flagged minor issues — sim-slice gap acknowledged in PR body, other items low impact.

PR #120, opened 2026-05-04
ContentRouter: force HLS transcode for codecs Apple TV cannot decode (MPEG-2, VC-1, VP9, AV1)

rrgomes:
## Summary

Apple TV silicon has no native decoder for several video codecs:

- **MPEG-2** — Apple removed hardware decode years ago.
- **VC-1 / WMV3** — undecodable on Apple silicon.
- **VP9** — no Apple TV chip has ever carried a VP9 hardware decoder.
- **AV1** — no Apple TV chip through A15 has it either; AV1 hardware
  decode landed on A17 Pro in 2023, but Apple has not yet shipped an
  Apple TV based on A17 Pro or later.

Today, a direct-play attempt against any of these surfaces **"Couldn't
Load Video"** and the user's only workaround is to manually dial down
their streaming-quality preset to force a transcode. Plex's first-party
tvOS client and Infuse handle this transparently — this PR makes
Rivulet do the same.

## What it does

- **`ContentRouter.swift`** — adds a `videoCodecsRequiringTranscode` set
  (`mpeg2`, `mpeg2video`, `mp2v`, `vc1`, `wmv3`, `vp9`, `av1`,
  normalized lowercased and stripped of hyphens/underscores).
  `ContentRouter.plan(for:)` short-circuits to `.hls` when the source
  video codec matches.
- **`PlexNetworkManager.buildHLSDirectPlayURL`** — extends the existing
  `forceVideoTranscode` parameter so that `directPlay`, `videoCodec`,
  and the Dolby Vision profile flags all flip to their transcode shapes
  (was only gating `directStream`). Plex now transcodes the asset to
  H.264 at the existing default ceilings.
- **`UniversalPlayerViewModel.startPlayback`** — forces the AVPlayer
  path for unsupported-codec content since `RivuletPlayer`'s pipeline
  can't consume the resulting HLS stream end-to-end (`loadHLS` only
  handles fallback against a pre-built transcode URL stored on
  `streamURL`, and the `.hls(url:)` route's URL carries just the server
  base).
- Both `buildHLSDirectPlayURL` callers in `UniversalPlayerViewModel`
  (the Rivulet HLS fallback builder and the audio-stream-switch
  rebuild) pass `forceVideoTranscode: ContentRouter
  .requiresVideoTranscode(metadata: metadata)`.

## Why this set of codecs

- **MPEG-2** is what initially surfaced this — older broadcast captures
  and classic-TV rips are frequently MPEG-2-in-MKV. Three test episodes
  from a Mayberry R.F.D. library (mpeg2video / ac3 / MKV) all
  reproduced "Couldn't Load Video" before this change and play
  correctly after it.
- **VC-1 / WMV3** are similarly undecodable on Apple silicon and worth
  pre-emptively covering — older WMV-derived encodes turn up
  occasionally in user libraries.
- **VP9** — Apple TV silicon has never carried a VP9 hardware decoder.
- **AV1** — exposed by an AV1+Opus-in-MKV YouTube-style title that
  reliably surfaced "Couldn't Load Video" before this change. Routing
  through forced HLS transcode resolves it: Plex transcodes video to
  h264 and audio to AAC and the resulting stream plays cleanly.

The set is conservative — codecs are added on confirmed real-world
failures, not speculatively, so we avoid forcing transcode of content
that would have direct-played fine.

## Test plan

- [x] Played three Mayberry R.F.D. episodes (S1/S2/S3) — all MKV /
      mpeg2video / ac3 — that previously failed; all play correctly via
      HLS transcode after this change.
- [x] Played a VC-1+DTS-MA test title; ContentRouter emitted
      `mkv | video=vc1 audio=dca-ma → HLS (codec requires transcode)`,
      playback succeeded.
- [x] Played an AV1+Opus test title (MKV) that previously failed.
      ContentRouter emitted `mkv | video=av1 audio=opus → HLS (codec
      requires transcode)`; `[VM] startPlayback`:
      `needsTranscodeForCodec=true mustUseAVPlayer=true`. Plays cleanly.
- [x] Played an h264+eac3 and an HEVC+FLAC title to confirm the
      direct-play / RivuletPlayer path is unchanged. ContentRouter
      routed to `LocalRemux` as expected; no regression.
- [x] Verified URL params: AV1/VP9 yield `directPlay=0 directStream=0
      videoCodec=h264`; H.264 keeps `directPlay=1 directStream=1
      videoCodec=h264,hevc`.

## Notes

### DIFF
--- a/Rivulet/Services/Plex/Playback/Pipeline/ContentRouter.swift
+++ b/Rivulet/Services/Plex/Playback/Pipeline/ContentRouter.swift
@@ -116,6 +116,21 @@ struct ContentRouter {
         "truehd", "mlp",                        // Dolby TrueHD / MLP
     ]
 
+    /// Video codecs Apple TV cannot decode natively at any tvOS version
+    /// (no hardware decoder, AVSampleBuffer rejects the format). Content
+    /// using these codecs must be server-side transcoded — direct-play
+    /// would fail with "Cannot Load Video".
+    ///
+    /// Stored normalized (lowercased, hyphens/underscores stripped); the
+    /// `requiresTranscode(videoCodec:)` helper applies the same normalization
+    /// before lookup.
+    static let videoCodecsRequiringTranscode: Set<String> = [
+        "mpeg2", "mpeg2video", "mp2v",          // MPEG-2 (broadcast captures, classic-TV rips)
+        "vc1", "wmv3",                          // VC-1 (older WMV-derived encodes)
+        "vp9",                                  // VP9 (no Apple TV hardware decoder)
+        "av1",                                  // AV1 (no Apple TV hardware decoder through A15 / 3rd-gen)
+    ]
+
     // MARK: - Route Decision
 
     /// Determine the primary playback route for the given content.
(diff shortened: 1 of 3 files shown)


l984-451 (approved):
Traced the full call chain: startPlayback gate routes unsupported codecs to AVPlayer regardless of the RPlayer toggle, prepareStreamURL builds the real transcode URL via buildRivuletHLSURL, buildHLSDirectPlayURL correctly flips directPlay=0/directStream=0/videoCodec=h264 and disables DV (transcoded output can't preserve it), and the audio-switch reload path is updated to match. Confirmed RPlayer's FFmpegDemuxer.createVideoFormatDescription only wires HEVC/H.264 — these codecs would throw unsupportedCodec immediately, so the AVPlayer-only enforcement is necessary, not redundant. Codec list (MPEG-2, VC-1, VP9, AV1) is accurate and conservative for the current pipeline; tracked the future work to add software video decode for these in Docs/FUTURE_WORK_RPLAYER_VIDEO_DECODE.md. LGTM.

PR #121, opened 2026-05-04
Player: surface timeline overlay on bare touchpad tap

rrgomes:
## Summary

A bare finger tap on the Siri Remote touchpad (no force/click) does nothing in the player today — the timeline / playback controls only appear when paused, when the user begins scrubbing, or when other state changes trigger them. Plex's first-party tvOS client surfaces controls briefly on tap, which is the discoverability-friendly default for "where am I in this episode?". This PR matches that behavior.

## What it does

- **`PlayerContainerViewController.swift`** — adds a `UITapGestureRecognizer` filtered to `.indirect` touch types (Siri Remote touchpad). On tap, calls the existing `viewModel.showControlsTemporarily()`, which already handles the show + auto-hide logic. Standard guards before triggering: bail if the info panel is open, the user is mid-scrub, post-video is showing, or playback has failed.
- The recognizer sets `allowedPressTypes = []` explicitly because `UITapGestureRecognizer` on tvOS defaults to waiting for a `.select` press in `allowedPressTypes`. Without overriding that, the recognizer never fires on a bare touch even when `allowedTouchTypes = [.indirect]` is set. Click-based interactions (play/pause via the `GCMicroGamepad` `buttonA` handler, dpad press handling) are unchanged.
- Coexists with the existing `UIPanGestureRecognizer` (swipe-to-scrub, swipe-down for info panel): the tap recognizes only on a quick touch-and-lift; if the user starts moving their finger, the pan recognizer takes over and the tap doesn't fire.

## Test plan

- [x] Bare tap during playback: timeline overlay appears, auto-hides ~3s after playback resumes. Verified on Apple TV 4K (tvOS 26.4).
- [x] Swipe-to-scrub (paused state) still works as before.
- [x] Touchpad click for play/pause still works as before.
- [x] Swipe-down to open the info panel still works.
- [x] Tap during scrubbing does not trigger overlay (gated by `isScrubbing` guard).
- [x] Tap with info panel open does not trigger overlay (gated by `showInfoPanel` guard).
- [x] Built clean against `upstream/main` with `CODE_SIGNING_ALLOWED=NO`.

## Notes

### DIFF
--- a/Rivulet/Views/Player/PlayerContainerViewController.swift
+++ b/Rivulet/Views/Player/PlayerContainerViewController.swift
@@ -20,6 +20,7 @@ class PlayerContainerViewController: UIViewController {
     private var hostingController: UIHostingController<AnyView>?
     private var cancellables = Set<AnyCancellable>()
     private var panGestureRecognizer: UIPanGestureRecognizer?
+    private var touchSurfaceTapGesture: UITapGestureRecognizer?
 
     // Directional gesture recognizers for IR remote support
     private var dPadLeftTapGesture: UITapGestureRecognizer?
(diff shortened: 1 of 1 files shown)


l984-451 (approved):
Verified showControlsTemporarily() does what the PR claims (auto-hide timer only fires during .playing), existing pan recognizer also uses allowedTouchTypes=[.indirect] so they coexist via UIKit's default tap/pan state-machine resolution, and click handling goes through GameController (buttonA.pressedChangedHandler) which is a completely separate event path from UIKit gestures — so allowedPressTypes=[] doesn't conflict. The tvOS UITapGestureRecognizer defaulting to allowedPressTypes=[.select] is the non-obvious gotcha that would trip anyone up; explicit override + clear comment is exactly right. LGTM.

PR #122, opened 2026-05-05
Playback: resume-or-restart prompt for in-progress items

rrgomes:
## Summary

- Pressing Play on an in-progress item now prompts "Resume from H:MM:SS" / "Start from Beginning" / "Cancel" instead of silently resuming. Matches the behavior of Plex's first-party tvOS app and Infuse.
- Covers the four detail-page Play sites (movie, episode, show/season next-up, episode card) and the direct-play paths from `PlexHomeView` and `PlexLibraryView` (Continue Watching primary press, hero carousel Play).
- Long-press "Watch from Beginning" paths are untouched — they already pass `fromBeginning: true` and bypass the new prompt.

## Implementation

- `PlexMetadata.formatResumeTime(_ ms:)` static helper — `H:MM:SS` when ≥ 1 hour, `M:SS` otherwise.
- `PlexDetailView.presentPlay(for:launch:)` centralises the prompt-or-launch decision; the four Play sites each pass an action closure.
- `playItemDirectly(_:fromBeginning:)` in `PlexHomeView` and `PlexLibraryView` is split into a public entry that prompts and a private `presentPlayerForItem(_:fromBeginning:)` that does the actual UIKit player presentation. The dialog's "Resume" / "Start from Beginning" buttons call back into the private impl so they don't re-trigger the prompt.
- The threshold for showing the prompt is the existing `PlexMetadata.isInProgress` (\`2% < progress < 90%\`), plus a defensive `viewOffset > 0` guard.

## Design note

This is a deliberate departure from the Apple-TV+-style "primary press always resumes silently" pattern Rivulet has today — restart-from-zero currently lives in `EpisodeContextMenuSheet` and `ContinueWatchingContextMenuModifier` long-press menus. Parity with stock Plex / Infuse felt worth the extra confirmation step (most folks coming to Rivulet are coming from those clients), but happy to gate this behind a setting if you'd rather preserve the silent-resume default. Let me know.

## Test plan

- [x] Press Play on an in-progress movie → dialog shows; both branches launch correctly.
- [x] Press Play on a fresh / never-watched movie → no dialog.
- [x] On a show detail page where the next-up episode is in progress → button reads "Continue S__E__"; press shows the dialog.
- [x] Press an in-progress episode card inside a season → dialog shows.
- [x] Home → Continue Watching → press an in-progress card → dialog shows.
- [x] Long-press an in-progress card → "Watch from Beginning" launches directly with no dialog (regression check).
- [ ] Library → hero carousel Play on an in-progress hero → dialog shows. *(Not exercised in local testing — same code path as Continue Watching, should behave identically.)*
- [ ] Shuffle Play (which sets \`playFromBeginning: true\` directly) still bypasses the dialog. *(Not exercised; code unchanged — Shuffle still passes the flag through.)*

---

Drafted with assistance from Claude (Anthropic). Every line was reviewed and tested locally on Apple TV before filing.

### DIFF
--- a/Rivulet/Models/Plex/PlexMetadata.swift
+++ b/Rivulet/Models/Plex/PlexMetadata.swift
@@ -542,6 +542,19 @@ extension PlexMetadata {
         return progress > 0.02 && progress < 0.9
     }
 
+    /// Format a Plex viewOffset (ms) for the resume prompt: H:MM:SS when ≥ 1 hour,
+    /// M:SS otherwise — matches stock Plex / Infuse "Resume from …" wording.
+    static func formatResumeTime(_ viewOffsetMs: Int) -> String {
+        let totalSeconds = max(0, viewOffsetMs) / 1000
+        let hours = totalSeconds / 3600
+        let minutes = (totalSeconds % 3600) / 60
+        let seconds = totalSeconds % 60
+        if hours > 0 {
+            return String(format: "%d:%02d:%02d", hours, minutes, seconds)
+        }
+        return String(format: "%d:%02d", minutes, seconds)
+    }
+
     /// Check if item should show as "watched" (completed, no active re-watch in progress)
     /// - Shows/Seasons: all episodes must be watched (viewedLeafCount >= leafCount)
     /// - Movies/Episodes: viewCount > 0 and not currently in progress
--- a/Rivulet/Views/Plex/PlexDetailView.swift
+++ b/Rivulet/Views/Plex/PlexDetailView.swift
@@ -94,6 +94,14 @@ struct PlexDetailView: View {
     @State private var playFromBeginning = false  // For "Play from Beginning" button
     @State private var isLoadingShufflePlay = false
 
+    // §26: Resume-or-restart prompt state. Shown when the user presses
+    // Play on an in-progress item (movie, episode, or show/season whose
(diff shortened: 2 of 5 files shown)


l984-451:
Yeah happy to add this feature, but lets gate it behind a setting defaulted to false/off like you suggest. I prefer the immediate resume and it matches what I see on Apple TV + app. Thanks!

rrgomes:
Updated to gate the prompt behind a Settings toggle, defaulted off — matches Apple TV+ behavior by default. The toggle lives in Settings → Playback as "Resume or Restart Prompt".

l984-451:
Thanks for iterating on the toggle gating — really appreciated. The post-rename refactor we landed earlier today moved the affected files (`Views/Plex/PlexDetailView.swift` → `Views/Media/MediaDetailView.swift` reparameterized on `MediaItem`, and `Views/Plex/PlexHomeView.swift` / `PlexLibraryView.swift` → `Views/Media/`), which would have made rebasing this PR a chunk of work for you on top of an architecture you haven't seen yet.

I ported your work directly onto current main in 507cb85 — same Settings toggle (default off), same three call sites gated, same dialog UX. Added a `SettingsDescriptors` entry per our convention so the description panel explains the behavior on focus, and dropped the `§26` markers from your internal planning doc. You're co-authored on the commit.

Closing this PR since the work has shipped on main. Thanks again!

PR #124, opened 2026-05-05
Player: error overlay buttons now visually focusable from Siri Remote

rrgomes:
## Summary

When playback fails, the "Couldn't Load Video" overlay rendered Try Again
and Dismiss as inline VStack buttons styled with `AppStoreButtonStyle`.
The buttons were technically reachable from the Siri Remote — touchpad
swipes moved focus between them, the focus engine activated them on
press, and `viewModel.retryPlayback()` / `viewModel.shouldDismiss = true`
fired correctly — but neither button visually highlighted when focused,
leaving the user unable to tell which button they were about to press.
The user-visible behavior was "the error screen has no focus and no way
out except menu/back," and the fact that focus *was* working was only
discoverable via diagnostic instrumentation.

## Root cause

`AppStoreButtonStyle` declares its own internal `@FocusState private
var isFocused: Bool` inside `makeBody(configuration:)` and binds
`.focused($isFocused)` to `configuration.label`. That binding doesn't
reliably reflect the button's true focus state when wrapped inside a
`ButtonStyle`, so the visual highlight (white-on-black + scale) never
engaged. Combined with `.focusEffectDisabled()` on the same view (which
suppresses the system-default lift effect), the buttons end up with no
visual indication when focused.

## Fix

Apply the standard tvOS overlay focus recipe used elsewhere in this
file (`WaitingToStartOverlay`, `HostDisconnectModal`):

- Lift `errorView`'s contents into a `PlayerErrorOverlay` struct so it
  can carry an outer `@FocusState` and per-button `.focused()`
  bindings.
- Wrap the call site in `.zIndex(50)` so the focus engine sees the
  error buttons as topmost in the player content layer.
- Apply `.focusSection()` + `.defaultFocus($focused, …)` inside the
  overlay's body to declare the focus container and default target
  declaratively.
- Claim focus from `.onAppear` with a one-tick delay
  (`asyncAfter(deadline: .now() + 0.05)`). Without the delay, the
  focus engine takes a tick to register the new focus container after
  the `playbackState=.failed` transition, so an immediate
  `@FocusState` assignment races and loses.
- Render Try Again / Dismiss inline (`Button + .buttonStyle(.plain) +
  .focusEffectDisabled`), driving the visual highlight off the outer
  `@FocusState` rather than `AppStoreButtonStyle`'s broken internal
  one. Visual styling matches `AppStoreButtonStyle` exactly
  (white-on-black + 1.1 scale on focus, white-on-glass when
  unfocused).

## Test plan

- [x] Try Again highlights white-on-black with a 1.1× scale on appear.
- [x] Touchpad swipes move the highlight between Try Again and
      Dismiss.
- [x] Try Again activates `retryPlayback()`; the overlay re-mounts
      after the retry re-fails and focus is reclaimed correctly.
- [x] Dismiss activates `shouldDismiss = true` and exits the player.

Repro: attempt to play a video whose source codec the device can't
decode. On a 2nd-gen Apple TV 4K (A12, no AV1 hardware decode), an
AV1-in-MKV video direct-plays and fails, surfacing "Couldn't Load
Video" cleanly. Note: #120 (also pending against this repo) routes
AV1 through HLS transcode, so once both PRs land you'll need a
different unsupported codec — or temporarily revert #120's routing
addition — to repro. Bringing down the Plex server to force a load
failure also reaches the `.failed` state, but loses surrounding view
state on the detail page (a separate pre-existing bug) and can mask
focus-related fixes — not recommended as a repro path.

## Notes

### DIFF
--- a/Rivulet/Views/Player/UniversalPlayerView.swift
+++ b/Rivulet/Views/Player/UniversalPlayerView.swift
@@ -919,9 +919,13 @@ struct UniversalPlayerView: View {
                 .zIndex(15)
             }
 
-            // Error State
+            // Error State. zIndex above the controls overlay so the
+            // SwiftUI focus engine sees the error buttons as topmost.
+            // The focus container itself (.focusSection + .defaultFocus
+            // + delayed onAppear claim) lives inside PlayerErrorOverlay.
             if case .failed(let error) = viewModel.playbackState {
                 errorView(error: error)
+                    .zIndex(50)
             }
 
             // Skip Button (intro/credits) - shows regardless of controls visibility, but not during post-video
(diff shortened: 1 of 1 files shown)


l984-451 (approved):
Verified the diagnosis: AppStoreButtonStyle's internal @FocusState/.focused() pair inside makeBody doesn't reliably reflect actual focus on tvOS — sibling style AppStoreActionButtonStyle in the same file explicitly documents this limitation. The bigger refactor (overlay struct + .focusSection + .defaultFocus + delayed onAppear claim) is the right call for an overlay context vs. just swapping the button style. Tracked a follow-up doc covering (a) the broken AppStoreButtonStyle now having zero call sites and being safe to delete, and (b) the inline actionButton helper diverging slightly from AppStoreActionButtonStyle's visual treatment. LGTM.

PR #139, opened 2026-05-10
AVPlayer/HLS: server-side subtitle burn-in (with 1080p cap)

rrgomes:
## Summary

The AVPlayer/HLS playback path now requests `subtitles=burn` so Plex
bakes the server-side-stored subtitle stream into the video frames.
On this branch HLG/DV content (routed here by #134) is the primary
use case; the user-forced "Apple Player" toggle also triggers it.
RivuletPlayer's HLS-fallback path is unchanged.

Stacks on **#134**. Mergeable without #134 — the user-forced
"Apple Player" toggle still triggers the path — but most of the
value lights up once #134 lands. (Diff in this PR includes #134's
commits until #134 merges.)

Three commits:

1. **`subtitles=burn` on the AVPlayer/HLS path** — adds a
   `burnInSubtitles` parameter to `buildHLSDirectPlayURL`; gates the
   in-playback subtitle picker on `inPlaybackSubtitleSwitchingSupported`
   (currently `rivuletPlayer != nil`). Mid-playback the SUBTITLES
   column shows an explanatory notice; focus skips to AUDIO. The
   pre-play subtitle picker on the detail page is unchanged.
2. **Active-track name in the in-playback notice** — closes a gap left
   by Plex's Now Playing dashboard, which surfaces only language +
   codec + SDH flag for burned-in subs (no track name).
3. **1080p cap on burn-in transcodes** — Plex's `subtitles=burn` filter
   chain forces an `hwdownload → inlineass → hwupload` round-trip per
   frame; 4K HEVC HLG bottlenecks even an M4 Pro transcoder below
   real-time speed. Capping at 1080p drops per-frame work ~4x.

## Why burn-in over a soft sub track

Plex's HLS subtitle pipeline + AVPlayer's WebVTT loader have three
structural limits on this path:

- **Single-stream manifest.** `subtitleStreamID` is session-scoped;
  the manifest emits exactly one `EXT-X-MEDIA TYPE=SUBTITLES` line.
  Mid-playback switching requires reissuing the whole transcode
  session.
- **First-cue threshold.** AVPlayer silently disables WebVTT
  renditions whose first cue is more than a few seconds into the
  timeline (common for cold opens). Working around it requires a
  local proxy that injects a synthetic threshold-satisfying cue.
- **Accessibility-styled cue background.** The synthetic cue still
  draws a brief opaque box at first appearance; AVPlayer applies the
  user's accessibility cue background as a separate compositing
  layer outside `::cue` CSS scope. No STYLE block or cue setting
  suppresses it.

Burning the subtitle stream into the video frames sidesteps all
three. Tradeoffs: no mid-playback Off / no mid-playback track
switching (pre-play picker still works); no client-side
accessibility styling on burned-in tracks.

## Future direction

The 1080p cap is a stopgap. The long-term fix for HDR Direct Play
is a custom Metal renderer in the spirit of Infuse — FFmpeg demux +
VideoToolbox HW decode + a `CAMetalLayer` with
`wantsExtendedDynamicRangeContent` and an HDR pixel format — which
bypasses Plex transcode entirely and removes both the resolution
cap and the burn-in dependency. Tracked separately as a possible
future direction; not part of this PR.

## Test plan

- [ ] Cold start an HLG title with embedded subtitles selected
      pre-play — subtitles appear immediately.
- [ ] Open the in-playback menu — SUBTITLES column shows
      "Currently playing: <track name>" + the explanatory notice;
      focus does not land in the column; right-arrow lands in AUDIO.
- [ ] Scrub forward 30+ minutes on a 4K HLG title — playback
      resumes within a few seconds. (Without the 1080p cap, the
      transcoder stalls for tens of seconds.)
- [ ] RivuletPlayer's HLS-fallback path — subtitles still arrive
      as a soft track (FFmpeg-decoded WebVTT). Pre-existing
      behavior; this PR should not regress it.

### DIFF
--- a/Rivulet/Services/Plex/Playback/Pipeline/ContentRouter.swift
+++ b/Rivulet/Services/Plex/Playback/Pipeline/ContentRouter.swift
@@ -176,18 +176,32 @@ struct ContentRouter {
             )
         }
 
-        // Apple TV has no native decoder for the source video codec
-        // (e.g. MPEG-2, VC-1, VP9, AV1). Direct-play would produce
-        // "Couldn't Load Video"; force a server-side transcode. The URL
-        // builder downstream picks up the same condition via
+        // Force a server-side transcode for content the rivuletPlayer
+        // pipeline can't render correctly. Two reasons:
+        //  • Apple TV has no native decoder for the source video codec
+        //    (MPEG-2, VC-1, VP9, AV1) — direct-play would produce
+        //    "Couldn't Load Video".
+        //  • Content uses HLG HDR — AVSampleBufferDisplayLayer on tvOS
+        //    accepts and decodes HLG frames without errors but renders
+        //    them black. tvOS's HLG→HDR10 internal conversion only fires
+        //    on the AVPlayer code path. Forcing a Plex transcode lands
+        //    H.264 SDR (or HDR10 if Plex's profile permits) on AVPlayer,
+        //    which the panel can actually display.
+        // The URL builder downstream picks up the same condition via
         // `forceVideoTranscode` so the HLS request is shaped as a real
         // transcode (directPlay=0, directStream=0, h264 target) rather
         // than a direct-play remux.
-        if let videoCodec = Self.primaryVideoCodec(from: context.metadata),
-           Self.requiresTranscode(videoCodec: videoCodec) {
-            reasoning.append("video_codec_requires_transcode_\(videoCodec.lowercased())")
+        if Self.requiresVideoTranscode(metadata: context.metadata) {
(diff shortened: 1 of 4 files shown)


l984-451:
Closing this. Burn-in is a workaround for an architectural limitation Rivulet is explicitly trying to avoid (the README's direct-play-no-transcoding stance), and silently downsampling 4K HDR to 1080p to make subtitles work isn't a tradeoff I want to ship even as a stopgap. The proper fix is local Metal HDR rendering with local subtitle compositing, which is what #142 is trying to extend to HLG. Let's focus the subtitle/HDR work there.

PR #142, opened 2026-05-11
HLG HDR: render via Metal sink for true Direct Play

rrgomes:
## Relationship to other open PRs

This supersedes #134's approach for HLG: instead of forcing server-side transcode to SDR, HLG content stays Direct Play via the Metal sink and reaches the panel as actual HDR. If this lands, #134 can be closed as obsolete. #139's server-side subtitle burn-in remains useful for other HLS-transcode paths and is unaffected.

## Why

AVSampleBufferDisplayLayer on tvOS accepts and decodes HLG frames
without errors but renders them black: AVSBL drives system tonemapping
from per-format-description color tags but does not expose the
macOS-equivalent CALayer.colorspace, so HLG output is not properly
configured. AVPlayer-based paths also fail (HEVC HLG rejected with
private CoreMediaErrorDomain -19601; HLS direct-stream and local-remux
fMP4 paths both surface the same error).

## What

Adds an alternate video sink — `MetalVideoRenderer` — that owns its own
`VTDecompressionSession` and presents decoded `CVPixelBuffer`s via
`CAMetalLayer` configured with:

  - `toneMapMode = .ifSupported`           (tvOS 18+)
  - `preferredDynamicRange = .high`         (tvOS 26+)
  - `contentsHeadroom` derived from HDR characteristics

When `ContentRouter.metalHLGSinkEnabled` is true (default) and
`DirectPlayPipeline` detects HLG in the first sample's
`CMFormatDescription`, it calls
`SampleBufferRenderer.enableMetalVideoSink()` which swaps the video
sink from AVSBL to the Metal renderer. The audio renderer stays
attached to the existing `AVSampleBufferRenderSynchronizer`; the Metal
renderer's `CADisplayLink` reads `synchronizer.currentTime()` to pick
which queued frame to present each tick.

The Metal renderer is **only** activated for HLG content. HEVC SDR,
AVC, DV PQ and all other paths continue to use AVSBL unchanged.

## Verified

On 3840x2160 HEVC HLG Main 10:

  - Plex server Now-Playing reports Direct Play (no transcode)
  - LG OLED panel reports HDR output mode
  - Audio + subtitles remain in sync
  - 25.2 fps present rate on 25 fps source; zero stutter
  - HEVC + AAC SDR content unaffected (still uses AVSBL)

### Backpressure

`isReadyForMoreMediaData` counts both queued frames AND in-flight VT
work (= enqueueCount − decodeCallbackCount) toward the depth cap.
Without this, the demuxer reads ahead and stuffs VT's pipeline,
callbacks complete in a burst, the drop-newest policy on the
decoded-frame queue silently discards excess, and the queue ends up
with gaps that double the effective pop interval — manifesting as
steady 12.5 fps stutter on 25 fps source. `decodeCallbackCount` is
incremented at the top of the callback (before the status check) so
failed decodes don't leak the in-flight count.

### Display pacing

The display loop pops already-due frames in a catch-up while-loop,
otherwise pops the head if its scheduled host time falls at or before
the next display vsync (`displayLink.targetTimestamp + 2ms`).
Per-branch counters (duePop / immPop / skipEmpty / skipFar +
farAvg / farMax) are logged every 5 s for future stutter-regression
diagnosis.

### HEVC parameter-set canonicalisation

FFmpeg's demuxer produces `hvcC` with `numArrays = 0` for some HEVC
content (typical of `hev1`-muxed streams tagged `hvc1`). AVSBL
tolerates this; standalone VT rejects with status -4 (`unimpErr`).
The renderer extracts VPS/SPS/PPS from the in-band NALU stream on the
first sample, rebuilds the format description via
`CMVideoFormatDescriptionCreateFromHEVCParameterSets`, and re-stamps
each subsequent sample via `CMSampleBufferCreateReady` with the
rebuilt FD (otherwise VT errors with
`kVTFormatDescriptionChangeNotSupportedErr` -12916).

### Pixel format

`VTDecompressionSession` defaults to `'p420'` (Apple compact 10-bit,
4 bytes per 3 pixels) for HEVC 10-bit content. Metal cannot sample
this with stock `r16Unorm`/`rg16Unorm` plane formats. The renderer
requests `'x420'` (`kCVPixelFormatType_420YpCbCr10BiPlanarVideoRange`)
explicitly via `imageBufferAttributes`.

## Feature flag

`ContentRouter.metalHLGSinkEnabled` is currently a `static let`
defaulting to true. If reviewers prefer a user-visible setting,
trivial to expose. If set to false, HLG content falls through to the
existing AVSBL path (renders black on tvOS, same as today).

### DIFF
--- a/Rivulet/Services/Plex/Playback/Pipeline/ContentRouter.swift
+++ b/Rivulet/Services/Plex/Playback/Pipeline/ContentRouter.swift
@@ -296,6 +296,13 @@ struct ContentRouter {
         return requiresTranscode(videoCodec: codec)
     }
 
+    /// Master switch for the Metal-renderer HLG direct play path. When true,
+    /// DirectPlayPipeline activates `SampleBufferRenderer.enableMetalVideoSink()`
+    /// on detecting HLG in the first sample's format description. When false,
+    /// HLG content falls through to the default AVSBL path (which renders
+    /// black on tvOS).
+    static let metalHLGSinkEnabled = true
+
     static func requiresTranscode(audioCodec: String) -> Bool {
         let normalized = audioCodec.lowercased()
             .replacingOccurrences(of: "-", with: "")
--- a/Rivulet/Services/Plex/Playback/Pipeline/DirectPlayPipeline.swift
+++ b/Rivulet/Services/Plex/Playback/Pipeline/DirectPlayPipeline.swift
@@ -1058,6 +1058,25 @@ final class DirectPlayPipeline {
             onError?(FFmpegError.noCodecParameters)
             return
         }
+        // Detect HLG in the actual decoded format description (more reliable
+        // than relying on Plex metadata for color characteristics). When the
+        // content is HLG and the Metal sink is enabled, switch the video sink
+        // from AVSampleBufferDisplayLayer to MetalVideoRenderer before
+        // configuring the layer color characteristics. AVSampleBufferDisplayLayer
+        // on tvOS renders HLG as black; the Metal sink uses CALayer.toneMapMode
+        // (tvOS 18+) + CALayer.preferredDynamicRange (tvOS 26+) to drive
+        // system tonemapping and emit true Direct Play HDR.
+        if ContentRouter.metalHLGSinkEnabled,
+           let ext = CMFormatDescriptionGetExtensions(videoFD) as? [CFString: Any],
+           let transfer = ext[kCMFormatDescriptionExtension_TransferFunction] as? String,
+           transfer == (kCMFormatDescriptionTransferFunction_ITU_R_2100_HLG as String) {
+            playerDebugLog("[DirectPlay] HLG detected in format description — enabling Metal video sink")
+            _ = renderer.enableMetalVideoSink()
+        }
+        // Hand the format description to the renderer so it can configure
+        // the active sink (AVSBL diagnostics, or Metal renderer VT session +
+        // layer color characteristics).
+        renderer.configureLayerForVideoFormat(videoFD)
         let audioFD = demuxer.audioFormatDescription
         let hasDV = demuxer.hasDolbyVision
         let activeAudioTrack = demuxer.audioTracks.first(where: { $0.streamIndex == demuxer.selectedAudioStream })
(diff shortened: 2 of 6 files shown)


l984-451:
Holding this for changes. The direction (Metal HDR sink with local compositing) is the right architectural answer for HLG and is what would also unblock proper subtitle rendering without the burn-in workaround from #139. But the current shape has issues that need addressing before merge:

1. VT decompression callback uses `passUnretained(self)` and `deinit` only calls `VTDecompressionSessionInvalidate(session)` without a preceding `VTDecompressionSessionWaitForAsynchronousFrames`. The session-invalidate is async; in-flight callbacks against a freed `self` are a use-after-free. The existing `flush()` correctly calls wait-for-frames before clearing; `deinit` needs the same treatment, or switch to `passRetained` with a matching `release` in the callback.

2. Pipeline state is built once with the layer's initial pixel format and never rebuilt on format change. A mid-playback HDR↔SDR transition (e.g. seeking across mixed content) will mismatch the shader output format and the layer; clear `pipelineState` in the format-change tear-down so it rebuilds against the new pixel format.

3. `Task { @MainActor in enqueueOnMain(sample) }` per sample creates ~60 task hops/sec on 60fps content, contending with UI work on MainActor. Move enqueue to a serial dispatch queue or actor that doesn't sit on MainActor.

4. `decodeCallbackCount` is declared `nonisolated(unsafe)` and is written from the VT callback thread but read from MainActor in `isReadyForMoreMediaData` (via `enqueueCount - decodeCallbackCount`). Use atomics or guard with `queueLock`.

5. `static let metalHLGSinkEnabled = true` defaults the new path ON with no user-recoverable kill switch. Convert to an AppStorage-backed toggle (default off would be safest while this is shipping) or at minimum make the flag mutable so a setting can override it.

Also: the BT.709/BT.2020 selector at the shader uses `pixelBufferIs10Bit(pixelBuffer)` as a proxy for color matrix. Wrong axis. SDR 10-bit HEVC exists and will get BT.2020 transforms applied to BT.709 content. The code already reads `kCMFormatDescriptionExtension_YCbCrMatrix` into a local `matrix` variable but doesn't actually use it. Wire that through to the shader uniform.

Nine em-dashes in new comments per CLAUDE.md, clean those up while you're in there.

Happy to revisit once these are addressed. The HLG case is currently covered by #134 (force HLS transcode), so there's no user-blocking pressure to land this quickly.

l984-451:
RPlayer is planned for retirement once Aether matures. HLG content currently routes to HLS transcode as a workaround, which is acceptable for the interim. Taking 1448 lines of custom Metal renderer + VTDecompressionSession plumbing for a player we're phasing out isn't worth it. Closing.

PR #152, opened 2026-05-21
PlexAuthManager: race connection probes instead of serial fallback

rrgomes:
## Summary

`findBestConnection` probes a server's advertised connections one at a time, and `connectionScore` sorts local addresses first. A client that is **not** on the server's LAN therefore tries every unreachable LAN address before reaching a routable one — and each unreachable address burns the full `testConnection` timeout (5 s) before the next is tried.

For a client off the server's LAN that is a multi-second stall on every server (re)connection, and a window during the stall in which an in-flight metadata request can land on a connection that is no longer valid and 404.

## Fix

Group a server's connections into priority tiers — local non-relay > remote non-relay > relay — and probe every candidate URL within a tier **concurrently** with a task group. The first working URL wins; the rest are cancelled. An unreachable tier now costs a single timeout instead of one-per-address compounded serially.

- The per-connection `http` → `plex.direct` → raw-HTTPS cascade and the cert-hash recovery are preserved, expressed as candidate variants (new `ConnectionCandidate` enum + `raceCandidates` / `resolveCandidate` helpers).
- `connectionScore` is removed — its tiering is now expressed directly by the tier filters.
- Probe timeout cut from 5 s to 3 s: `/identity` is a tiny response a reachable server answers well within that, and a tighter cap bounds how long an unreachable address can hold up a race.

This is the connection-racing pattern Plex's own first-party clients use.

## Test plan

- [x] Built for tvOS, Debug.
- [x] Device on the server's LAN — local tier resolves immediately; behavior unchanged.
- [x] Device off the server's LAN — the multi-second "Connecting" stall before reaching a routable address is gone.
- [x] Switching between a local server and a remote server resolves promptly in both directions.
- [x] Relay-only reachability still falls through to the relay tier.

### DIFF
--- a/Rivulet/Services/Plex/PlexAuthManager.swift
+++ b/Rivulet/Services/Plex/PlexAuthManager.swift
@@ -231,145 +231,135 @@ class PlexAuthManager: ObservableObject {
         return await task.value
     }
 
-    /// Find the best working connection for a server
-    /// Priority depends on server.httpsRequired:
-    /// - If httpsRequired=true + local: plex.direct (valid TLS cert) > HTTP > HTTPS
-    /// - If httpsRequired=false + local: HTTP (fastest) > plex.direct
-    /// - Remote: current behavior (plex.direct URLs from API)
+    /// A single connection candidate to probe when resolving a server.
+    private enum ConnectionCandidate: Sendable {
+        /// Probe this URL directly.
+        case plain(String)
+        /// Probe a raw-HTTPS URL; if it fails with a TLS error that
+        /// reveals the server's plex.direct hash, fall back to the
+        /// matching plex.direct URL.
+        case httpsWithCert(uri: String, address: String, port: Int)
+    }
+
+    /// Find the best working connection for a server.
+    ///
+    /// Connections are grouped into priority tiers — local non-relay >
+    /// remote non-relay > relay — and every candidate URL within a tier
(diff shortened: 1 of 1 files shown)


l984-451:
Closing this. #145 (already merged) is the narrower correctness fix that solves the user-visible server-selection identity mismatch bug. The broader connection-racing refactor here introduces non-determinism that I think outweighs the per-request latency improvement: within a tier, all URL variants race together, so on healthy networks the persisted `selectedServerURL` flaps between `http://192.168.x.x` and `https://x-x-x-x.{hash}.plex.direct` based on which TLS handshake wins the coin flip. The cached URL feeds AVPlayer cert validation, library fetches, and reconnection logic, so flapping can manifest as "playback works yesterday, fails today" with no code change between sessions.

If you want to take another swing at this, the shape I'd want is "connections in tier order, variants in parallel within each connection." That preserves the implicit preference ordering (original `connection.uri` over derived plex.direct forms when `httpsRequired=false`) while still capturing the parallelism win.

Separately: the 5s → 3s `timeoutInterval` cut in three places needs its own justification. The racing approach already addresses the per-address-stack stall problem the description cites, so the timeout cut is a separate decision with separate risk. 3s is aggressive on consumer WiFi where `/identity` legitimately takes 3-4s under load.

PR #158, opened 2026-05-27
Collections: Library/Collections picker + per-item collection footer

rrgomes:
## Summary

Adds Collections as a first-class browse surface, with no behavior
change for libraries that have no collections.

- **Library detail view** gains a centered Library / Collections tab
  picker at the top. The Collections tab renders a grid of the
  section's collections (composite poster + name + child count, e.g.
  "57 Movies"); selecting one pushes a `CollectionDetailView` listing
  the collection's items in release-date order. The picker only
  renders the Collections tab when the library actually has any, and
  per-library picker state is cached so re-entering a library
  restores the last-used tab.

- **Item detail view** gains a "<Collection Name> Collection" footer
  row on movies that belong to a Plex collection — other items in
  the collection, release-date order, excluding the current item.

- **New Settings toggle "Related Row on Item Pages"** (default on) to
  suppress the existing provider-suggested "Related" footer
  independently of the new collection footer. Wired through
  `SettingsDescriptors` so it lands in the sidebar like every other
  setting.

## Why three commits

Split for review:

1. `Collections: preserve collection IDs through MediaItemDetail` —
   data-model groundwork. `PlexMediaMapper.detail` was dropping
   `Collection[].id` and keeping only the tag name, so
   `collectionItems(matching:in:)` had no way to resolve a name back
   to an id without an extra round-trip (the reason
   `PlexProvider.collectionItems` stubbed to `[]`). Adds a
   `MediaCollectionRef` value type and a
   `PlexNetworkManager.getCollections(sectionId:)` listing endpoint.

2. `Collections: Library/Collections picker on library detail view` —
   the picker + grid + `CollectionDetailView`. Star Wars'
   long-standing broken composite cache on PMS is worked around by
   rewriting each collection thumb URL's
   `/composite/<timestamp>` segment to `/composite/0` so Plex
   regenerates from current collection state and re-caches.

3. `Collections: per-item "<name> Collection" footer + Related Row toggle` —
   the per-item footer, plus the Settings toggle to suppress the
   existing Related row.

## Test plan

- [ ] Open a library that has collections — picker appears, Library
      tab is default, Collections tab shows the section's collections
      with poster + name + child count.
- [ ] Switch to Collections, tap a collection — pushes the items list
      in release-date order.
- [ ] Re-enter the library — last-used tab is restored.
- [ ] Open a library with **no** collections — picker is absent;
      behavior is the existing hero / hubs / grid.
- [ ] Open a movie that belongs to a collection — footer "<Name>
      Collection" row appears with sibling items in release-date
      order.
- [ ] Open a movie that does **not** belong to a collection — no
      collection footer; Related row still appears.
- [ ] Settings → toggle "Related Row on Item Pages" off — Related
      row disappears on the next detail view; collection footer is
      unaffected.
- [ ] Star Wars library: collection composite thumbs load (the
      `/composite/0` rewrite).

### DIFF
--- a/Rivulet/Models/Media/MediaCollectionRef.swift
+++ b/Rivulet/Models/Media/MediaCollectionRef.swift
@@ -5,7 +5,7 @@
 //  Lightweight reference to a collection an item belongs to. Carries the
 //  provider-native id plus the display name. Distinct from a full collection
 //  entity (which a future browse surface would model as `MediaItem` with
-//  `kind == .collection`) — this type only needs to identify and label a
+//  `kind == .collection`); this type only needs to identify and label a
 //  collection, not render it as a tile.
 //
 
--- a/Rivulet/Models/Media/MediaCollectionRef.swift
+++ b/Rivulet/Models/Media/MediaCollectionRef.swift
@@ -1,0 +1,19 @@
+//
+//  MediaCollectionRef.swift
+//  Rivulet
+//
+//  Lightweight reference to a collection an item belongs to. Carries the
+//  provider-native id plus the display name. Distinct from a full collection
+//  entity (which a future browse surface would model as `MediaItem` with
+//  `kind == .collection`) — this type only needs to identify and label a
+//  collection, not render it as a tile.
+//
+
+import Foundation
+
+struct MediaCollectionRef: Hashable, Identifiable, Sendable {
+    /// Provider-native ID. For Plex this is the stringified `Collection[].id`
+    /// used by `/library/sections/{sectionId}/all?collection={id}`.
+    let id: String
+    let name: String
+}
--- a/Rivulet/Models/Media/MediaItemDetail.swift
+++ b/Rivulet/Models/Media/MediaItemDetail.swift
@@ -27,4 +27,5 @@ struct MediaItemDetail: Sendable {
     // Wave 1 additions for the detail view
     let nextEpisode: MediaItem?      // shows only — Plex `OnDeck`, Jellyfin `/Shows/NextUp`
     let collections: [MediaCollectionRef]  // collections this item is tagged with
+    let librarySectionId: String?    // for resolving collection items via section-scoped API
 }
--- a/Rivulet/Models/Media/MediaItemDetail.swift
+++ b/Rivulet/Models/Media/MediaItemDetail.swift
@@ -26,5 +26,5 @@ struct MediaItemDetail: Sendable {
 
     // Wave 1 additions for the detail view
     let nextEpisode: MediaItem?      // shows only — Plex `OnDeck`, Jellyfin `/Shows/NextUp`
-    let collections: [String]        // collection names this item is tagged with
+    let collections: [MediaCollectionRef]  // collections this item is tagged with
 }
--- a/Rivulet/Services/MediaProvider/Plex/PlexMediaMapper.swift
+++ b/Rivulet/Services/MediaProvider/Plex/PlexMediaMapper.swift
@@ -336,7 +336,8 @@ enum PlexMediaMapper {
             contentRating: meta.contentRating,
             rating: meta.rating,
             nextEpisode: nextEpisode,
-            collections: collections
+            collections: collections,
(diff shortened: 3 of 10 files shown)


l984-451:
Holding this for changes before merge. A few things:

1. `PlexLibraryView.swift:1542` references `WatchTogetherController.shared`, which doesn't exist on main. Looks like a bad rebase carry-over. Won't compile as-is.

2. `CollectionDetailView.loadItems()` uses `getChildren(ratingKey:)`, which returns Plex's section default order (alphabetical titleSort). The PR description and per-item footer both promise release-date order. Switch to `getCollectionItems(sectionId:collectionId:sort:"originallyAvailableAt")` (the method this PR already adds), passing `librarySectionID` + the collection id from `coll`.

3. Five em-dashes in new comments. Per CLAUDE.md these are a hard no in code/docs:
   - `Models/Media/MediaCollectionRef.swift:14`
   - `PlexLibraryView.swift` diff lines 504, 611, 725, 762

4. The preview-overlay sub-item dismissal change in `PlexLibraryView` (`onSubItemNavigation` block, ~lines 400-417) is unrelated to Collections. Please split it into its own PR or drop it — main already has the canonical fix via the `PreviewContainerViewController.dismissTopmost(then:)` helper (commit bf1ed7d), and bundling unrelated changes here will make this harder to bisect later.

Once those are addressed I'll come back and review again. The Collections browse surface itself is well-shaped and Library/Collections picker behavior is the right call. Thanks for the work on this.

rrgomes:
Pushed force; rebased onto current upstream/main and addressed all four items.

1. `WatchTogetherController.shared`: dropped. The player launch site at `PlexLibraryView.swift` matches main now.

2. `CollectionDetailView.loadItems`: switched to `getCollectionItems(sectionId:collectionId:sort:"originallyAvailableAt")`. To make `coll.librarySectionID` populated for callers downstream, `getCollections` now backfills it from its `sectionId` input (Plex's `/library/sections/{sectionId}/collections` responses don't carry it at the item level).

3. Em-dashes: dropped from all five flagged sites.

4. Preview-overlay sub-item dismissal: dropped. Main's canonical `dismissTopmost(then:)` from bf1ed7d is in use.

Ready for re-review.

l984-451:
I like this concept and functionally its great, but we need to be more careful with the UI. The buttons cover the library hero and are an extra down click to get through content. I think the more cohesive and better UX implementation is to replicate the home screen carousel item, when going from the expanded view down to the details view. The season pills do not show until navigating down, so the focus goes straight to episodes, and then you can focus back up on the season pills if desired. Then when navigating back up, the season pills fade out. We should replicate that effect here.
Also, when switching between Library and Collections buttons, the buttons overlap for a split second during the switch. This doesn't look good. Should probably just add a bit more spacing between them.

rrgomes:
Reworked per your review. Two changes:

(1) Carousel-style pill reveal that mirrors the season pills on a show page: the pills are hidden while the library hero owns focus (pressing Down from the hero skips straight to content), revealed once focus moves into the content rows, and fade back out on return to the hero. While revealed in content they sit recessed (dimmed, with a subtle active indicator) so the active tab doesn't read as focused; they brighten to the full white active fill only when the picker itself is focused. The hero now renders in both Library and Collections modes so the picker behaves identically in either.

(2) Widened the pill spacing (24 -> 40) so the focus scale-up no longer overlaps the neighbouring pill during a switch.

One note: this picker-over-hero interaction is easy to miss because the Library Hero is off by default (Settings -> Appearance), in which case the picker just stays put as the top-of-screen nav. With the hero on, the reveal relies on the hero keeping initial focus, which is one of the fixes in a companion PR (#170) for pre-existing library-hero bugs (backdrop clamp, initial focus, entry flash, land-on-Play). The two were developed together and both touch PlexLibraryView, so they'll have a trivial overlap in the onHeroFocused closure if both land.

l984-451:
@rrgomes I'm going to close this since its all based on the old SwiftUI views. If you have something updated for UIKit please open another one, or I'll work on it myself when I have time.

PR #161, opened 2026-05-27
Library item titles + detail-page action-row scroll

rrgomes:
## Summary

Two independent UI improvements bundled because they share files.

### 1. Library item titles

Library grid (`PlexLibraryView`) and home-screen `InfiniteContentRow`
tiles (Recently Added etc.) now show the item's title under each
poster, with the release year as a subtitle for movies. Continue
Watching cards keep their existing self-contained layout.

This matches Plex's own tvOS client default — title under every
poster, year under movie posters.

Implementation notes:

- The whole tile (poster + label) scales uniformly on focus via a
  per-tile `@FocusState` and outer `scaleEffect`, with the inner
  `MediaPosterCard` hover-highlight suppressed so the label is never
  covered by the expanded poster.
- Label is clamped to `ScaledDimensions.posterWidth` so long titles
  truncate within poster width rather than stretching the tile.
  `LazyHStack` doesn't enforce a per-item width the way `LazyVGrid`'s
  columns do — that's why the library grid was already truncating
  correctly and Home rows weren't.

### 2. Detail-page action-row horizontal scroll

`MediaDetailView`'s `actionButtons` HStack was overflowing on items
with many buttons (e.g. movies that also surface a Trailer button):
the Play button collapsed and wrapped its time text ("43m" → "43"
and "m" on separate lines), and the sibling circle-to-pill buttons
got clipped on both sides when focused.

Two changes:

- `playButtonLabel` uses `lineLimit(1).fixedSize()` on the time text
  and `fixedSize(horizontal: true, vertical: false)` on its HStack,
  so the play button holds its intrinsic width regardless of what a
  sibling does on focus.
- The `actionButtons` HStack is wrapped in
  `ScrollView(.horizontal, showsIndicators: false).scrollClipDisabled()`.
  When the row exceeds available width, tvOS's focus engine
  auto-scrolls the focused button into view; `scrollClipDisabled`
  keeps the focus-driven pill expansion from being clipped at the
  viewport edge. When the row fits naturally there's no scrolling and
  the layout is identical to before.

## Test plan

- [ ] Home (Continue Watching + Recently Added rows): poster tiles
      show title underneath; movie tiles show year as subtitle.
- [ ] Library grid: same labels; long titles truncate within poster
      width (try "4 Artists Paint 1 Tree: A Walt Disney 'Adventure
      in Art'" in a Shorts library if you have one).
- [ ] Focus a tile — the whole tile (poster + label) scales together;
      label never gets covered by the expanding poster.
- [ ] Continue Watching cards retain their existing self-contained
      layout (no double labels).
- [ ] Detail page for a movie with a Trailer button — action row
      fits naturally; Play button's time text doesn't wrap.
- [ ] Detail page for a movie with many extra buttons — the action
      row scrolls horizontally as focus moves; pill expansion isn't
      clipped at the edge.

### DIFF
--- a/Rivulet/Views/Media/MediaDetailView.swift
+++ b/Rivulet/Views/Media/MediaDetailView.swift
@@ -1063,13 +1063,24 @@ struct MediaDetailView: View {
     }
 
     private var actionButtons: some View {
-        HStack(spacing: 18) {
-            if isPlayableFromProvider {
-                playableActionButtons
-            } else {
-                watchlistActionButton
+        // ScrollView so the row keeps each button at its intrinsic width
+        // (play button + multiple circle-to-pill buttons) when the total
+        // exceeds the available width — happens on items with an extra
+        // optional button (e.g. Trailer) or when a focused button
+        // expands into a pill. tvOS focus engine auto-scrolls the
+        // focused button into view; .scrollClipDisabled lets the
+        // focus-driven pill expansion render past the viewport edge
+        // without clipping.
+        ScrollView(.horizontal, showsIndicators: false) {
+            HStack(spacing: 18) {
+                if isPlayableFromProvider {
+                    playableActionButtons
+                } else {
+                    watchlistActionButton
+                }
             }
         }
+        .scrollClipDisabled()
         .disabled(!allowActionRowInteraction)
     }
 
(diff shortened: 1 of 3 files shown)


l984-451:
I cherry-picked #2 fix, but have specifically removed titles from past versions. I prefer the titleless asthetic, and it matches Apple TV+ style. We can gate this behind another setting toggle if you want, though lets be mindful of creating too many options for users to choose.

PR #163, opened 2026-05-27
Library Item Titles: opt-in toggle

rrgomes:
## Summary

Re-files #161 with an opt-in toggle instead of unconditional titles.

Adds a "Library Item Titles" toggle (default OFF) that renders each
poster tile's title underneath, with the release year as a subtitle
for movies. Affects the library grid (PlexLibraryView) and home-screen
InfiniteContentRow tiles (Recently Added, etc.). Continue Watching
cards keep their existing self-contained layout.

When off, the original poster-only layout is preserved exactly,
including MediaPosterCard's native hover-highlight focus effect. When
on, the whole tile (poster + label) scales uniformly on focus via a
per-tile `@FocusState` and outer `scaleEffect`, with the inner
hover-highlight suppressed so the label is never covered by the
expanded poster.

## Context

Originally implemented with a toggle but stripped before upload to
avoid toggle creep; on reflection, and given your invitation in #161,
re-shipping it gated makes sense. Default OFF preserves the titleless
aesthetic for new installs; users who prefer a Plex-tvOS-style
labeled layout can opt in from Settings.

## Open questions for review

1. **Related row.** The toggle currently covers Library + Home rows.
   Should it also affect the Related row, where you just dropped
   poster titles via `f0d0e76`? Happy to either extend the toggle
   there or leave Related titleless regardless. Your call.

2. **Collections detail (#158).** Once #158 merges, I'd like to file
   a small follow-up gating `CollectionDetailView`'s per-tile labels
   behind the same toggle so the surfaces stay coherent. Including it
   in this PR would create a cross-PR dependency, so it lives as a
   follow-up.

## Test plan

- [ ] Settings shows "Library Item Titles" toggle, default OFF.
- [ ] Toggle OFF: library grid + home rows render poster-only (matches
      pre-#161 main exactly).
- [ ] Toggle ON: each tile shows title under poster; movie tiles show
      year under title; show/season tiles show title only.
- [ ] Focus scale-up: entire tile (poster + label) scales together;
      label is not covered by expanded poster.
- [ ] Continue Watching cards unchanged regardless of toggle state.

### DIFF
--- a/Rivulet/Views/Media/PlexHomeView.swift
+++ b/Rivulet/Views/Media/PlexHomeView.swift
@@ -1207,6 +1207,7 @@ struct InfiniteContentRow: View {
     @State private var hasReachedEnd = false
     @State private var totalSize: Int?
     @FocusState private var focusedItemId: String?  // Track which item is focused (format: "context:itemId")
+    @AppStorage("showLibraryItemTitles") private var showLibraryItemTitles = false
 
     /// Create a unique focus ID for an item in this row
     private func focusId(for item: PlexMetadata) -> String {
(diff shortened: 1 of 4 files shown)


l984-451:
Found a layout bug while testing.

Long titles aren't constrained to the poster width, so they spill horizontally and stretch the column/row layout. See screenshot: "Happiness Is a Warm Blanket, Charlie Brown" (2011) blows out the first tile's footprint, throwing off the whole row's alignment.

`.lineLimit(1)` prevents vertical wrapping but doesn't truncate width when there's no horizontal constraint.

Please look for edge cases like these in the future.

rrgomes:
Good catch — fixed in the new push.

The label VStack had `.frame(height: 44, alignment: .top)` but no width constraint, so `.lineLimit(1)` prevented vertical wrapping without bounding horizontal extent — long titles like "Happiness Is a Warm Blanket, Charlie Brown" spilled into the next column and blew out the row's alignment.

Constrained the label to `ScaledDimensions.posterWidth * scale` (same width as the poster column) and added explicit `truncationMode(.tail)` so long names truncate with an ellipsis at the poster's right edge.

The equivalent width clamp was already in `PlexHomeView`'s `InfiniteContentRow` from the original implementation — Library grid was the one we missed when re-spreading the toggle pattern. Will run the long-title check on more PRs going forward.

l984-451:
same as 158 due to the UIKit migration

PR #167, opened 2026-05-28
Sidebar: allow tab swap from detail views and Settings sub-pages

rrgomes:
## Summary

Selecting a sidebar tab while a detail view (NavigationStack push) or a Settings sub-page was on screen used to be a no-op. The sidebar still surfaced via touchpad swipe-left, but taps had no effect.

Three cooperating causes, all addressed:

1. `toolbarVisibility(.hidden, for: .tabBar)` was applied while `nestedNavState.isNested` or `isSettingsSubPage` was true. With the tab bar hidden, tvOS still lets touchpad swipe-left surface the sidebar visually — but the surface is non-interactive, so taps don't reach the binding setter.
2. Even if a tap had reached the setter, `tabSelection.set` short-circuited via `guard !isNested, !isSettingsSubPage else { return }`.
3. The deeper structural issue: tvOS locks the visible tab content to its NavigationStack while the stack is at depth ≥ 1. With (1) and (2) fixed, the setter fires and `selectedTab` writes to `@State` successfully, but the displayed tab doesn't follow until the push is popped.

## Fix

- Drop the guard and the `isNested`/`isSettingsSubPage` parts of the toolbar-hide condition.
- Add `NestedNavigationState.popToRootSubject` (Combine `PassthroughSubject`). When the sidebar setter fires from a nested state, broadcast pop-to-root.
- Each per-tab content view (`PlexHomeView`, `PlexLibraryView`, `PlexSearchView`, `MusicHomeView`, `SettingsView`) subscribes via `.onReceive` and clears its detail/sub-page `@State` on receipt.
- Defer the `selectedTab` write to the next runloop via `Task { @MainActor in }` so SwiftUI processes the pop first and then sees a clean root-level tab swap.

Music library views keep `toolbarVisibility(.hidden)` — they own the whole screen and have their own internal layout.

Menu/back from a Settings sub-page still triggers `goBack()` as before; in-Settings back navigation is unaffected.

## Test plan

- [ ] From Home, open a movie detail → swipe-left → tap a library → detail dismisses, library root appears.
- [ ] Same from show, episode, extra detail pages.
- [ ] From a Settings sub-page → swipe-left → tap a tab → sub-page pops, tab switches.
- [ ] Settings sub-page back navigation (menu/back) still pops one level at a time.
- [ ] Normal root-level tab swap unchanged.
- [ ] Returning to a library after tab-swapping from its detail lands at the library root, not the prior detail page (intentional — abandoned detail).

### DIFF
--- a/Rivulet/Views/Media/PlexHomeView.swift
+++ b/Rivulet/Views/Media/PlexHomeView.swift
@@ -239,6 +239,14 @@ struct PlexHomeView: View {
                     presentPreview(request: request)
                 }
             }
+            .onReceive(nestedNavState.popToRootSubject) { _ in
+                // Sidebar tab swap from inside a detail push: clear our
+                // nav-stack @State so the push pops to root, unblocking
+                // the tab swap. See NestedNavigationState.popToRootSubject.
+                selectedItem = nil
+                selectedMusicItem = nil
+                pendingPreviewNavigation = nil
+            }
         }
         .onChange(of: selectedItem) { _, newValue in
             print("[PlexHome] selectedItem changed: \(newValue?.title ?? "nil") (itemID: \(newValue?.ref.itemID ?? "nil"))")
--- a/Rivulet/Views/Media/PlexLibraryView.swift
+++ b/Rivulet/Views/Media/PlexLibraryView.swift
@@ -313,6 +313,13 @@ struct PlexLibraryView: View {
                     presentPreview(request: request)
                 }
             }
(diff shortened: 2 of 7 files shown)


l984-451:
The UIKit home opens detail via PreviewCarouselViewController as a modal -- not a NavigationStack push -- so isNested never goes true for the primary detail path and the sidebar tab swap works correctly. Settings sub-page navigation was addressed separately. No cherry-pick needed. Closing.

PR #168, opened 2026-05-28
Cast & Crew row: render real headshots and make it D-pad navigable

rrgomes:
## Summary

The Cast & Crew row on a media detail page was broken in two visible ways. After this PR it shows real cast/crew headshots and is reachable by the focus engine.

## Issue 1: Every person showed a generic icon

`PlexMediaMapper.detail` and `PlexMediaMapper.artworkURL` unconditionally prepended `serverURL` and appended `X-Plex-Token` to every thumb path. That works for server-relative thumbs (e.g. `/library/metadata/123/thumb/456`), but cast/crew thumbs come back from Plex as fully-qualified URLs into the public metadata CDN:

```
thumb=https://metadata-static.plex.tv/a/people/<hash>.jpg
```

The composed URL ended up like `https://<server>https://metadata-static.plex.tv/...?X-Plex-Token=…`, the request failed, and `PersonCard` fell back to its `person.fill` placeholder — for every actor. Detect `http://`/`https://` paths in `artworkURL` and return them unchanged; route the inline `personURL` closure in `detail()` through the helper so the same handling applies to cast, directors, writers, and chapter thumbs.

## Issue 2: Focus engine couldn't enter the row

`CastCrewRow` wrapped each card in `Button { } label: …` with no `@FocusState`/`.focused` tracking and no focus-driven visual effect. The focus engine never traversed into the section, so the row wasn't reachable by D-pad and its horizontal `ScrollView` couldn't be scrolled to reveal additional cast members.

Mirror the existing pattern from the detail-page Related row (`MediaItemAgnosticRow`): a local `@FocusState focusedID: String?`, `.focused($focusedID, equals: id)` on each button, and a 1.10× scale-up when focused. Empty button actions are kept — there's no person-detail destination yet — but the buttons are now valid focus targets, so the engine traverses into the row.

## Test plan

- [ ] Open any movie detail page → scroll down to Cast & Crew → real headshots render (no generic person icon for actors with a CDN thumb; placeholder still appears for entries Plex didn't supply a thumb for).
- [ ] D-pad down lands on a person card; the focused card scales up.
- [ ] D-pad right scrolls the row horizontally, revealing additional cast members.
- [ ] Posters, backdrops, and other artwork still render normally (server-relative paths unaffected by `artworkURL`'s pass-through branch).

### DIFF
--- a/Rivulet/Services/MediaProvider/Plex/PlexMediaMapper.swift
+++ b/Rivulet/Services/MediaProvider/Plex/PlexMediaMapper.swift
@@ -51,8 +51,20 @@ enum PlexMediaMapper {
     // MARK: - Artwork
 
     /// Helper for building Plex artwork URLs from arbitrary paths.
+    ///
+    /// Most thumb/art fields come back as server-relative paths (e.g.
+    /// `/library/metadata/123/thumb/456`) that need the server origin
+    /// prepended and an `X-Plex-Token` appended. Cast/crew thumbs and
+    /// some other people-related artwork instead come back as fully
+    /// qualified URLs into Plex's public metadata CDN (e.g.
+    /// `https://metadata-static.plex.tv/.../people/<hash>.jpg`). Return
+    /// those unchanged — prepending the server origin produces a
+    /// malformed URL whose request fails, falling back to a placeholder.
     static func artworkURL(_ path: String?, serverURL: String, authToken: String) -> URL? {
         guard let path else { return nil }
+        if path.hasPrefix("http://") || path.hasPrefix("https://") {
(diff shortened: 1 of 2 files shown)


l984-451:
Both fixes are already in main. The artworkURL absolute-URL passthrough (Issue 1) is present in PlexMediaMapper.artworkURL at the hasPrefix check, and personURL in detail() has the same guard inline -- picked up when building the UIKit person detail page. The CastCrewRow SwiftUI focus fixes (Issue 2) don't apply; the UIKit detail page has its own cast row with native UIKit focus. Nothing to cherry-pick. Closing.

PR #170, opened 2026-05-29
Library hero: clamp oversized backdrop art, keep initial focus, fix entry flash, land on Play

rrgomes:
## Summary

Four pre-existing fixes to the library hero, each independent of any
one library. They only surface with the Library Hero enabled (Settings
→ Appearance), which is off by default, so they are easy to miss.

- **Backdrop art no longer balloons the layout.** `HeroBackdropImage`
  rendered its art with `.aspectRatio(.fill)` but no bounding frame, so
  an item whose source art has extreme dimensions (e.g. a near-square
  title card used as a backdrop) reported the scaled image size as its
  own layout size, ballooning the content stack to thousands of points
  and, via the center-aligned state ZStack, shoving the hero a full
  screen off the top (posters appeared to cover the hero). The image is
  now anchored to a flexible `Color.clear` and clipped, so it fills its
  container regardless of source dimensions.

- **Hero keeps initial focus.** On freshly-loaded libraries the grid's
  `ensureInitialFocusIfNeeded` pulled focus into the poster grid,
  burying the hero. It now stands down while the hero is on screen, so
  the hero's Play button keeps the focus it takes on entry.

- **No cold-entry flash.** The instant hero pick during library-switch
  restoration used the item list while the eventual pick used the
  promoted hub, so the hero visibly swapped a beat after entry. The
  instant pick now prefers the same hub-based pick, and the hero is
  chosen once and not re-picked when hubs change.

- **Entering the hero lands on Play.** tvOS directional focus landed on
  whichever button sat above the column focus came up from (often the
  "next" arrow); the hero now re-anchors to its Play button on entry.
  Only fires on entry, so navigating between the hero's buttons
  afterwards is unaffected.

Developed alongside the Collections picker work in #158, which builds
its picker-reveal-over-hero behavior on the "keep initial focus" fix
here. Both touch `PlexLibraryView` and will have a trivial overlap in
the `onHeroFocused` closure if both land.

## Test plan

(requires Settings → Appearance → Library Hero on)

- [ ] Open a library whose hero art is off-shape (very tall/wide or
      near-square) — the hero fills its band and is not pushed
      off-screen.
- [ ] Open a freshly-loaded library — focus lands on the hero's Play
      button, not the poster grid below.
- [ ] Switch between libraries quickly — the hero does not visibly swap
      to a different item a beat after entry.
- [ ] Move focus onto the hero from the rows below — it lands on Play,
      then arrow-navigates to the other buttons normally.
- [ ] Library Hero off (default) — no behavior change.

### DIFF
--- a/Rivulet/Views/Media/Hero/HeroOverlayContent.swift
+++ b/Rivulet/Views/Media/Hero/HeroOverlayContent.swift
@@ -131,6 +131,15 @@ struct HeroOverlayContent: View {
         .onChange(of: focusedButton) { oldButton, newButton in
             if newButton != nil && oldButton == nil {
                 onHeroFocused?()
+                // Always land on Play when entering the hero. tvOS's directional
+                // focus otherwise lands on whichever button sits above the
+                // column you came up from (often the "next" arrow); Play is the
+                // hero's primary action, so re-anchor to it for a predictable
+                // entry. Only fires on entry (oldButton == nil), so navigating
+                // between buttons once inside is unaffected.
+                if newButton != .play {
+                    focusedButton = .play
+                }
             } else if newButton == nil && oldButton != nil {
                 onHeroExited?()
             }
--- a/Rivulet/Views/Media/HeroBackdropSupport.swift
+++ b/Rivulet/Views/Media/HeroBackdropSupport.swift
@@ -237,9 +237,21 @@ struct HeroBackdropImage<Placeholder: View>: View {
     }
 
     private func imageView(_ image: UIImage) -> some View {
-        Image(uiImage: image)
-            .resizable()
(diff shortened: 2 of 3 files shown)


l984-451:
All four fixes here address the deprecated SwiftUI `PlexLibraryView` / `HeroOverlayContent` / `HeroBackdropImage` path. The UIKit home (PlexHomeViewController + HeroBackdropView + HeroOverlayCell) was written from scratch and doesn't have these bugs by construction -- the backdrop balloon issue doesn't exist because UIImageView with scaleAspectFill + clipsToBounds is inherently layout-safe, and focus management is handled differently in the UIKit home. No cherry-pick needed. Closing.

PR #172, opened 2026-05-30
Detail metadata: variant-accurate audio, resolution, and full episode air date

rrgomes:
Enriches the hero metadata badge row on the detail page so it reflects the actual variant being played, across every video kind (it rides the shared `MediaDetailView` row).

- **Audio** — codec + channel layout for the default track, e.g. `E-AC3 5.1`, `DTS-HD MA 7.1`, `TrueHD 7.1 Atmos`, `AAC 2.0`, `FLAC 1.0`. The codec name distinguishes lossless families from lossy so the label is self-describing; Atmos is read from Plex's stream `profile` (object metadata), not the channel count. Stereo/mono are shown as `2.0`/`1.0` rather than dropped.
- **Resolution** — derived from Plex's computed `videoResolution` (correct for cropped widescreen whose pixel height is below nominal), falling back to track height. Interlaced sources get an `i` suffix (`1080i`/`576i`/`480i`) from the per-stream scan type; 4K/720 have no interlaced broadcast form and stay progressive; an absent scan type is treated as progressive so nothing is mislabeled.
- **Air date** — episodes show the full original air date (`Sep 24, 2019`); movies keep year-only.

Adds `MediaSource.videoResolution` and `AudioTrack.qualityLabel`; `VideoTrack` carries `scanType`; `MediaItemDetail` carries `originallyAvailableAt`. The quality-badge unit tests are rewritten for the new scheme (including interlaced cases).

Verified on device against a real Plex library. Note for review: this touches `MediaDetailView.swift`, also modified by #158 — the change here is confined to the hero metadata row, so the two should rebase independently.

### DIFF
--- a/Rivulet/Models/Media/AudioTrack.swift
+++ b/Rivulet/Models/Media/AudioTrack.swift
@@ -10,9 +10,11 @@ import Foundation
 struct AudioTrack: Hashable, Sendable, Identifiable {
     let id: String
     let index: Int                 // stream index in the container (AVPlayer track index)
-    let codec: String              // "eac3", "dts", "truehd", "aac", "opus"
-    let channels: Int?             // 2, 6, 8
-    let channelLayout: String?     // "5.1", "7.1", "Atmos" if present
+    let codec: String              // "eac3", "dca", "truehd", "aac", "opus"
+    let profile: String?           // Plex codec profile: "ma" (DTS-HD MA),
+                                   // "dolby truehd + dolby atmos", etc.
+    let channels: Int?             // 1, 2, 6, 8
+    let channelLayout: String?     // Plex layout, e.g. "5.1(side)", "stereo"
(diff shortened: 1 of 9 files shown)


l984-451:
Thanks for this. We cherry-picked the model-layer improvements into our UIKit implementation: AudioTrack gains `profile` and a `qualityLabel` computed var (e.g. "E-AC3 5.1", "DTS-HD MA 7.1", "TrueHD 7.1 Atmos"), VideoTrack gains `scanType`/`isInterlaced`, and MediaSource gains `videoResolution` with an updated `qualityBadges()` that prefers the provider-computed label and handles interlaced suffixes (480i/576i/1080i). The UIKit chrome also picks up the episode full air date. The SwiftUI MediaDetailView changes weren't pulled since that view is deprecated in our UIKit migration.

PR #173, opened 2026-05-30
Detail page: make the hero description a focusable element that opens the full summary

rrgomes:
## Summary

Replaces the "(i) More Info" action button with a focusable hero description: the synopsis/tagline block itself becomes the focus target, and selecting it opens the full text in `SummarySheet`. This unifies movies, shows, and episodes on one focusable-summary pattern and matches the Apple-TV/Infuse idiom of a silent primary action. It supersedes the recently-added info-circle button — the `showSummarySheet`/`SummarySheet` plumbing is reused, just triggered from the description instead of a separate button.

## What changes

- **Focusable synopsis.** The whole synopsis/tagline block is one focus target; clicking it presents `SummarySheet`. The visible text is full-width, but the focus target is a narrow transparent proxy pinned over the Play column — tvOS resolves a directional Down by geometry and won't consume an `onMoveCommand` for a section-less leaf, so a full-width focus frame would steal Down onto the centered Watched circle. Narrowing the frame leaves only Play beneath it.
- **Movie layout.** Movies now show the tagline as a headline with the synopsis beneath (previously a tagline hid the synopsis entirely): unwatched → tagline + spoiler-blurred synopsis; watched → tagline + un-blurred; no tagline → the synopsis takes the headline slot.
- **Preview overlay.** The description is focusable both in the standalone detail view and in the preview overlay; in the overlay, Up from the synopsis collapses back to the item carousel, so one Up from the buttons reaches the synopsis and a second reaches the carousel.
- **Focus cue.** A leading accent bar plus brightening and a soft glow, with the system button highlight disabled so it never washes out the text.
- Restyles the `SummarySheet` Done button to read clearly when focused.

Verified on device. Note for review: touches `MediaDetailView.swift`, also modified by #158 and #172 — this change is confined to the hero description block and action row, so they rebase independently.

### DIFF
--- a/Rivulet/Views/Media/MediaDetailView.swift
+++ b/Rivulet/Views/Media/MediaDetailView.swift
@@ -435,6 +435,12 @@ struct MediaDetailView: View {
         .onChange(of: nextUpEpisode) { oldValue, newValue in
             let ref = currentItem.ref.itemID
             previewFocusLog.info("[NextUp] changed: \(oldValue?.ref.itemID ?? "nil", privacy: .public) -> \(newValue?.ref.itemID ?? "nil", privacy: .public) isExpandedPreviewFlow=\(isExpandedPreviewFlow) showExpandedChrome=\(showExpandedChrome) currentFocus=\(String(describing: focusedActionButton), privacy: .public) ref=\(ref, privacy: .public)")
+            // Standalone detail view: once Play enables, promote focus from the
+            // shuffle fallback claimed on appear (for shows whose nextUp had not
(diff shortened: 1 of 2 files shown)


l984-451:
The synopsis-expansion feature is already implemented in the UIKit detail page, but it lives in the below-fold About card rather than the hero chrome. Selecting the left card in the About section opens the full text via InfoPopupViewController. The hero description in our implementation is display-only (not a focus target), which matches the simpler approach we've settled on for the UIKit surface. Closing since the capability is covered, just accessed differently.

PR #174, opened 2026-05-30
Detail page: surface a browsable Extras row with per-extra detail pages

rrgomes:
## Summary

Plex items carry an Extras array — trailers, featurettes, behind-the-scenes, deleted scenes, interviews — that the detail page previously reduced to a single trailer URL. This surfaces the whole list as a first-class browsable row, with each extra opening its own detail page.

## What's included

- **Extras row.** A horizontal row of extra tiles beneath Recommendations. New `MediaExtra` model + `ExtraSubtype` enum (canonical Plex categories, fixed render order); `MediaItemDetail` now carries the full extras array.
- **Pressing a tile opens its detail page.** Matching every other poster in the app (library grids, Recommended, Collections) rather than playing it directly. The extra maps through the normal single-video path (Plex `type=="clip"` classified as `.movie`), so it gets the full action row for free; a fresh detail view is pushed so the below-fold content belongs to the extra.
- **Per-season extras.** On a show page, the selected season's extras render in a second row ("<Season> Extras") beneath the show-level row, which retitles to "Show Extras" when both are visible. Collapses when a season has none.
- **Extra detail page.** Uses the extra's own full title and backdrop, and labels the type by its real subtype ("Trailer", "Featurette", …) instead of "Movie".
- **Library files only.** Drops Plex-supplied IVA (online) extra streams so the row surfaces only real library files.
- **Trailer-button precision.** The header Trailer button derived its URL from the first extra matching `subtype=="trailer"` OR `type=="clip"` — and since every Plex extra is `type=="clip"`, that matched any extra. With the full list now surfaced this would point the button at an arbitrary featurette. Drop the clip clause and gate on a loose title match, so the button points at a real trailer for the feature or doesn't render.

One new file (`MediaExtra.swift`); the rest integrates into the existing detail view and Plex mapper. Verified on device.

Note for review: touches `MediaDetailView.swift` and the Plex mapper, which other open PRs (#158, #168, #172, #173) also modify — the changes here are scoped to the Extras row, the extra-detail path, and the trailer-URL derivation, so they rebase independently.

### DIFF
--- a/Rivulet/Models/Media/MediaExtra.swift
+++ b/Rivulet/Models/Media/MediaExtra.swift
@@ -1,0 +1,72 @@
+//
+//  MediaExtra.swift
+//  Rivulet
+//
+//  Agnostic representation of a bonus-feature item attached to a movie,
+//  show, season, or episode. Provider-mapped from `PlexExtra` today;
+//  carries enough to render a 16:9 tile + play through the universal
+//  player by ratingKey.
+//
+
+import Foundation
+
+struct MediaExtra: Identifiable, Hashable, Sendable {
+    /// Provider-native ID (Plex ratingKey).
+    let id: String
+    let title: String
+    let subtype: ExtraSubtype
+    let durationSec: Int?
+    let thumbURL: URL?
+}
+
+/// Canonical Plex extra categories. Render order is the declaration
+/// order below.
+enum ExtraSubtype: Int, CaseIterable, Sendable, Comparable {
+    case trailer
+    case featurette
+    case behindTheScenes
+    case deletedScene
+    case interview
+    case scene
+    case short
+    case unknown
+
+    /// Map Plex's raw `subtype` / `extraType` strings to the enum.
+    /// Plex uses `sceneOrSample` interchangeably with `deletedScene` on
+    /// some library types; both map to `.deletedScene`.
+    static func fromPlex(subtype: String?, extraType: Int?) -> ExtraSubtype {
+        switch subtype {
+        case "trailer": return .trailer
+        case "featurette": return .featurette
+        case "behindTheScenes": return .behindTheScenes
+        case "deletedScene", "sceneOrSample": return .deletedScene
+        case "interview": return .interview
+        case "scene": return .scene
+        case "short": return .short
+        default:
+            // Fallback to extraType when subtype is absent/unrecognised.
+            // Plex uses extraType=1 for trailer; other integer codes are
+            // inconsistent across versions, so don't over-map.
+            if extraType == 1 { return .trailer }
+            return .unknown
+        }
+    }
+
+    static func < (lhs: ExtraSubtype, rhs: ExtraSubtype) -> Bool {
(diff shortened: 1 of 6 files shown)


l984-451:
@rrgomes Hey, this isn't specific to this PR, but I wanted to let you know I'm working on a large code refactor. Essentially SwiftUI is not very performant on tvOS, and I'm converting the more complicated screens to use UIKit. Super frustrating to be spending time on this refactor, but it will be worth it. 
That being said, you should hold off on UI additions based on the current SwiftUI architecture because a lot of it will have to be refactored or wrapped in some way to work with the UIKit code once its in place. 
I'm hoping to have most of the work done by end of this week, maybe sooner. 
Thanks again for your contributions!

rrgomes:
Thanks for the heads-up; the refactor makes good sense.

I'll hold off upstreaming further UI changes until your refactor lands, so you're not reviewing things that have to be re-done.

One question on the PRs already open (#142, #158, #163, #167, #168, #170, #172, #173, #174): would you prefer to land any of them before the refactor, or should I hold and re-file against the UIKit architecture once it's in? A few are mostly additive (the HLG Metal sink in #142 is player-side; the rest are detail/library UI), but you'll know which will survive the conversion. I can rebase whatever's worth keeping onto the new structure when it's ready.

Thanks for Rivulet and for being receptive to my contributions. I happened upon it when I was looking for a good base on which to build a tvOS Plex client that would improve or fix many of the things that have always bothered me about Plex's own client, many of them trivial to fix but never done because they're focused on their "new Plex Experience" and other things I don't care about at all.

My biggest motivation was to develop a client that could properly handle Watch Together, a feature I rely on heavily and which Plex is dropping from its rebuilt client apps. I have a very functional implementation working now that I've been using for WT sessions with friends. I haven't upstreamed any of that yet because it's a big piece of work and I'm not sure you'll even want it. It's built on Plezy's relay, with additive extensions, so it stays protocol-compatible with theirs.

After I started work on Rivulet I discovered that Plezy has since released a tvOS version of their app. I test-drove it. It's capable and it includes some features I've already added to Rivulet and others that I plan to add, but it's awkward and jerky and feels like a port of an app built for another platform rather than a proper tvOS app. And their HDR and DV implementation doesn't seem right because my OLED panel doesn't recognize Plezy's output as HDR or DV. So I'd prefer to keep making Rivulet better.

My goal is to make Rivulet the only tvOS client for Plex that I'll need.

l984-451:
@rrgomes Finally done with the migration. I'm going to start working through PRs again. Thanks for your patience and your contributions!

l984-451:
Thanks for this! We've manually implemented parts of this PR into the UIKit detail page: the `ExtraSubtype` enum with per-subtype SF Symbol icons, and the extras row itself. We chose not to bring in the per-extra detail page navigation — in Rivulet, tapping an extra plays it directly rather than opening a detail view. The `hasLocalFile` concept is preserved on the model for potential future use but isn't active yet since the Plex metadata API doesn't return `Media[]` on extras. Appreciate the contribution!

PR #177, opened 2026-06-11
fix(persistence): recover from corrupt store instead of crashing at launch

Artic0din:
## What was broken

`RivuletApp.sharedModelContainer` built the SwiftData `ModelContainer` with a single `try ModelContainer(...)` whose only `catch` was `fatalError("Could not create ModelContainer: ...")`. This runs at launch in the `@main App`. A corrupt store, full disk, or failed migration made init throw, turning every launch into a permanent crash with no path to the UI to reset — unusable until the app was reinstalled, which loses all local watch progress. This is audit finding **C-1**.

## What this fixes

Replaces the `fatalError` with a tiered fallback in a new testable `ModelContainerFactory`:

1. Open the persistent store as-is.
2. On failure: log to Sentry, delete `default.store` and its `-shm`/`-wal` siblings, retry once.
3. Still failing: fall back to an in-memory store so the app stays usable for the session.

`fatalError` now only fires if even the in-memory store cannot be created. `ContentView` shows a dismissable non-fatal banner when local history was reset or kept only in memory. The `@Model` types stay registered (hide path), so no drop-entities migration (C-2) is needed.

## Test plan

- `xcodebuild build -scheme Rivulet -destination 'platform=tvOS Simulator,name=Apple TV 4K (3rd generation)'` passes.
- New `ModelContainerFactoryTests` (7 cases) cover all three tiers, the reset-throws path, the exhausted-throws path, and store-file removal. All pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased]
+
+### Fixed
+
+- Launch no longer crashes when the local SwiftData store is corrupt, un-migratable, or unwritable. The store is reset and rebuilt, falling back to an in-memory store as a last resort, with a dismissable on-screen notice when local watch history was reset (audit finding C-1).
+
 ## 1.0.0 (Build 48)
 
 - Added Discover + Watchlist tabs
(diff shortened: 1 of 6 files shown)


Artic0din:
Opened against the wrong repo by mistake — reopening against the Artic0din fork.

PR #185, opened 2026-06-25
System subtitle settings for Rivulet Player (claude)

tresby:
Done via claude.

Rivulet Player now gets its font/size/background/colour/outline settings from the Apple TV system settings.

Also adjusted background to be tighter around the text to be closer to the native apple subtitles.


l984-451:
Merged manually into main as 52de6cb with two fixes applied: `CaptionStyle` marked `@unchecked Sendable` for Swift 6, and the uniform outline ZStack reduced from 8 copies to 4 diagonal corners. Thanks @tresby!

PR #209, opened 2026-07-17
Sub improvements/Aether bump/intro animation/dot timer

tresby:
Builds off the dot timer PR.

Adds:
- Subtitles now will pass along colour info
- Subtitle delay can be adjusted per movie/episode/channel
- Subtile height can be adjusted universally
- Webvtt subs don't blink via custom subtitle renderer
- Skip Markers now use a fill animation vs countdown 
- Bumped Aether to 5.5.0 (better Aus tv live support/dvb_teletext)

Hopefully I've left info.plst alone now - I feel like it was recently working on build 66 or the commits just after (pre 67).

This should make the dot timer PR obsolete as this should include all of that.


l984-451:
thanks! And you're saying PR208 should just be ignored/removed?

tresby:
yep this includes everything in there I believe.

Couple of small things I need to fix up but once it passes your review and if you decide to implement ill fix them up.

l984-451:
Cool. Thanks for all this! I went through it and, because there were so many changes in one PR, had to pull it apart and implement them myself with some changes.

**What I've Merged:**

Subtitle content colors, following the system "Video Override" caption setting, plus the fix to match cues on the clock's source axis.
Live TV WebVTT captions rendered through our own overlay so rolling captions stop flickering.
Per-item subtitle Delay and global Height steppers on both the VOD and Live TV rails.
The skip button fill sweep replacing the "· N" countdown suffix (debating about this. May undo).
Live TV Menu closing one layer at a time.
The community marker database (IntroDB) backfill.

**What I changed on the way in:**

IntroDB is now opt-in (a Playback setting, off by default) and gates the network lookup itself. It sends only the show's ID and episode number, and it is confidence-gated.
I dropped the second database host. The public API only documents one, and the other 404s.
Menu handling in the Live TV player uses the same mechanism the VOD player already uses for the parallel system dismiss, rather than a new path.

**What I left out:**

The hero auto-advance dot countdown. It moves the focused item out from under the user while the Play button is focused, so pressing Play can start the wrong title.  I need to think more about this. Auto advancing feels expected on a hero, but when focus is on the hero (which is always) you don't want it to auto advance from under you.
Skip Preview. Nothing produces a "preview" marker. Plex doesn't emit the type and the IntroDB API only returns intro, recap, and outro, so the setting and its plumbing were dead code.
This last one I haven't spent too much time reviewing myself, but I'll let Claude explain: The loadLive change for the multiview grid slots. This one is worth explaining. It bundled two things: a 30-minute disk-backed DVR rewind window, and a switch to playing HLS channels through AVPlayer directly instead of the engine. The DVR window is pure cost on a muted preview tile that has no scrubbing, times up to four slots, so that part is a clear no. The direct-HLS switch is a real tradeoff: it's plausibly lighter for four concurrent streams, but it moves the grid's mute-across-slots logic onto a different player backend, and the current path already handles auth-header IPTV channels because the engine does the authenticated fetch itself. The grid works today, and I'd rather not add a second playback path for a niche opt-in feature on spec. Is the performance issue an real thing you've experienced? If so please submit the HLS features again separate from the DVR rewind and an explanation of the performance issues.

tresby:
Thanks for reviewing those.

Yep fair point on the hero timer. I feel like I do like the look of the effect but agree it may not add much more than that.

On skip preview - I could’ve sworn one of the dbs had previews too albeit not common. Maybe the implementation didn’t happen correctly - I will look into it, though for the only show I can think of with previews is maybe just Succession?

Menu handling being the same for Live TV as VOD makes perfect sense and I had noticed it was different but hadn’t addressed that yet - unsure why Claude went down the path of a new system.

On the multiview edit - I’m not 100% sure on this one, must have been something Claude tackled without me picking up on it because I hadn’t really explored Multiview via Aether for Live TV yet.

Hope that wasn’t too much to untangle.

PR #219, opened 2026-07-19
Codex/plex live tv playback fixes

tresby:
- Live TV guide lazy loading (wasn't implemented)

- Added Live TV OSD progress bar (Left of screen is start of program, indicator is current time, right of screen is end of program) note: happy to change this to something more suitable should live tv seek be added

- Fixed quick guide scrolling could cause channel logos to appear in the wrong sports

- Fixed first channel guide cell offset

- Adjusted day header on EPG to be central

- DVB_teletext will use page 801 if Apple TV is set to region AU so all DVB_teletext subs display (default is subtitle which should work okay for most regions - can add more region specifics if needed)

- DVB_teletext wiring added to VOD too (was only live tv before)

- Fixed Plex guide data image handling

- Fixed plex live tv request logic to be compatible with Aether (direct stream) + dashboard reporting - note: this may need improving on - I personally still load via m3u due to plex not dealing with dvb_teletext well, but that may just be a limitation of plex based live tv


tresby:
Sorry Bain, kept making changes that overlapped so became quite a large PR again. 

Hopefully not too hard to untangled. I know some multi view stuff was touched - but again unsure how relevant that is as we will probably need to re-write stuff to suit Aether multi view if it's possible.

I don't personally use Plex Live TV - but my change to Aether for Live TV broke Plex based Live TV - so I figured id better get it working. If there is any improvements you want to make to it, by all means adjust what and how it requests the channels.

Live TV progress bar is tied to the current program for now, potentially would be changed when seeking is added incase we tune a HLS playlist with a large window, it may have the ability to seek back past the window in the progress bar.

The DVB_teletext changes are just setting txt_page to 801 if the Apple TV is set to AU region. I can't imagine people are consuming too much teletext subs from outside their region, so I figured this would be a safe way to implement while keeping the default as subtitle - which should suffice for most regions, I think AUS teletext is just bit finicky from what I've read in the past.

l984-451:
Thanks for this one, a lot of it shipped in 1.0.4 (69): the guide lazy loading, the channel logo fix, the guide layout and artwork fixes, the program progress bar, and the teletext handling. The live tv handshake was reworked in the same direction as yours, including your session identifier and profile encoding, and it should fix cloud guide DVB channels failing to start. Since you run that setup, can you test build 69 and let me know? I'll close this once it's confirmed working.

PR #220, opened 2026-07-20
fix: scale poster proportionally when opening a title

### LINKED ISSUE #221: Poster stretches instead of scaling when opening a title
### Before submitting

- [x] I have searched for existing issues

### Describe the bug

Opening a movie or show from a Home row stretches the poster instead of scaling it. The tile snapshot is animated straight to the carousel card frame, and because the two frames have very different shapes the artwork visibly squashes on the way in.

### Why it happens

Tapping a tile captures `cell.snapshotView(afterScreenUpdates:)` and flies it into the card in `PreviewCarouselViewController.runRowEntryMorph()`. A snapshot view is bitmap-backed, so its contents stretch to fill whatever bounds it's given:

- poster tile **296 × 444** (2:3) — `MediaRowMetrics`
- centered card **1744 × 1028** (~1.70:1) at 1920 × 1080 — `PreviewCarouselGeometry.centeredCardFrame`

That's **×5.89 horizontally against ×2.32 vertically**, so the aspect ratio drifts from 0.667 to 1.697 across the morph, peaking at **2.54× distortion**. Dismiss has the same problem, stated outright as `contentMode = .scaleToFill`.

### Steps to reproduce

1. From Home, tap any movie or show tile in a row (not hero Info or the tile menu's "More Info" — those open already-expanded and don't morph)
2. Watch the poster as it grows into the full-screen card

Most visible on posters with a strong circular or facial element.

### Expected

The poster should scale proportionally and stay anchored to the top, letting the overflow run off the bottom and crop, the way the Apple TV+ app does it.

### Proposed fix

Draft PR:
- #220

Adds a clipping window that takes the card frame while the snapshot inside is aspect-filled and top-anchored, so it only ever scales uniformly. Same construction as `BackdropPlaneView.Panel`.

> **Caveat:** the geometry is unit-tested (21 standalone assertions, verified against the shipped function), but the change has **not been run on a device or in the simulator** and nobody has watched the animation yet. The PR is deliberately a draft. It also notes a follow-up worth a design call: the poster now ends at ×5.89 while the backdrop plane cross-fades in at ~×1, which could read as a pop on titles that have no backdrop art.

bhargav95:
Fixes https://github.com/l984-451/Rivulet/issues/221

Opening a movie or show stretched its poster instead of scaling it. This makes the growth proportional and anchors it to the top, the way the Apple TV+ app does it.

## The bug

Tapping a tile captures `cell.snapshotView(afterScreenUpdates:)` for each visible tile and flies it into the carousel card in `runRowEntryMorph()`. A snapshot view is bitmap-backed — its contents stretch to whatever bounds you assign — and the two frames are far apart in shape:

- poster tile **296 × 444** (2:3) — `MediaRowMetrics`
- centered card **1744 × 1028** (~1.70:1) at 1920 × 1080 — `PreviewCarouselGeometry.centeredCardFrame`

That's **×5.89 across against ×2.32 down**, applied to a bitmap. The dismiss direction had the same problem, stated outright as `contentMode = .scaleToFill`.

The `standaloneDetail` paths (hero Info, tile-menu "More Info") don't show it — they open already-expanded with no card-grow morph.

## Before (Zendaya's head is squished, change playback speed if the effect is too fast) 

https://github.com/user-attachments/assets/335c0eab-c4de-49cf-b69c-019586ebb2e3 

## After (scale like Apple TV+ ) 
https://github.com/user-attachments/assets/f31ce61d-3593-4211-82c9-b98fb5ecd5bd

## The fix

`EntryMorphWindowView` is a clipping window that takes the frame the snapshot used to take. The snapshot lives inside it, aspect-filled and top-anchored, so it only ever scales uniformly and overflows the bottom edge — which is then cropped. Same construction as `BackdropPlaneView.Panel` (clipping container over an oversized image).

Three call sites move from `snapshotView.frame = …` to `window.setWindow(…)`: snapshot setup, the entry morph, and the exit morph.

No `CADisplayLink` is needed, unlike the corner-radius lerp. Both endpoint content frames satisfy `h = w / aspect`, and linear interpolation preserves that:

```
h(t) = h₀ + t(h₁ − h₀) = (w₀ + t(w₁ − w₀)) / aspect = w(t) / aspect
```

Top-anchoring is free for the same reason — content `y` is 0 at both endpoints, so it's 0 throughout.

Aspect is read from `sourceFrame` rather than hardcoded, so Continue Watching's landscape tiles (357 × 277) need no special case.

## Testing

**Unit** — `EntryMorphWindowViewTests.swift`, **12 tests, all pass** in the tvOS Simulator: both endpoints, uniform scale on both axes, top-anchoring, bottom overflow, landscape tiles, the exit direction, degenerate input, a swept-t proportionality check, and a reproduction of the old distorting path.

```bash
xcodebuild test -scheme Rivulet -destination 'platform=tvOS Simulator,name=Apple TV' \
  -only-testing:RivuletTests/EntryMorphWindowViewTests
```

**Visual** — built and run in the tvOS Simulator; the morph was watched on screen and recorded. Before/after clips above.

## Notes from self-review

Worth a look before this merges:

1. **Scale discontinuity at the hand-off (most likely to need a design call).** The morph now ends with the poster at ×5.89 while `backdropPlane` cross-fades in at roughly ×1. For items *with* backdrop art the fade covers it, since they're different images. For items *without* it, `BackdropPlaneView.loadArtwork` falls back to `item.artwork.poster` (`BackdropPlaneView.swift:164`) and renders it `.scaleAspectFill` over the full stage — so the fade blends the same poster against itself at a ~6× scale difference. The old code didn't have this, because its end state was squashed but still roughly ×1. **Worth testing a title with no backdrop specifically.**
2. **Dismiss during the entry morph.** Pressing Menu inside the 0.45s window is reachable (`.entryMorph` returns `.dismissOverlay`), and the entry animator's completion sets `entrySnapshotContainer.isHidden = true`, blanking the exit morph mid-flight. Pre-existing, not introduced here, but it lives in a function this PR rewrites.
3. `contentFrame` is pure `CGGeometry` but inherits `@MainActor` from the class, which forces the test class to be `@MainActor` too. Could be `nonisolated`, or lifted next to `PreviewCarouselGeometry`.
4. `EntryMorphWindowView.init` sets its frame twice — `super.init(frame:)` then `setWindow` with the same size. Taking the full source `CGRect` (both callers have it) would set it once.
5. `test_setWindow_movesWindowAndRefitsContent` uses exact `CGSize` equality; it passes only because 296/444 round-trips exactly in double precision. Every other numeric assertion in the file uses `accuracy: 0.5` and this one should too.

Changelog line added under `1.0.3 (68)`.

Happy to fold any of the above in, or drop the exit-morph half if you'd rather keep dismiss exactly as it behaves today.

### DIFF
--- a/Rivulet/Views/Components/WhatsNewView.swift
+++ b/Rivulet/Views/Components/WhatsNewView.swift
@@ -143,6 +143,7 @@ struct WhatsNewView: View {
         ("1.0.4 (69)", [
             "New Content Filtering in Playback settings can mute strong language and skip scenes during playback, with per category controls and a quick toggle in the player",
             "Language muting works on any title with subtitles, and scene skipping uses filter lists in the MCF or EDL format from a URL you provide, without ever changing your files",
+            "Opening a movie or show now grows its poster proportionally instead of stretching it out of shape",
             "The backdrop that appears while paused now waits ten seconds instead of five, so you have more time to look at the paused frame",
             "Pressing Back while the paused backdrop is showing now brings back the paused frame instead of closing the player",
             "Fixed playback getting stuck after leaving the app and coming back, the video now picks up right where you left off",
--- a/Rivulet/Views/Media/PreviewCarousel/UIKit/PreviewCarouselViewController.swift
+++ b/Rivulet/Views/Media/PreviewCarousel/UIKit/PreviewCarouselViewController.swift
@@ -56,6 +56,12 @@ final class PreviewCarouselViewController: UIViewController {
     private var hasPreparedEntrySnapshots = false
     private var didStandaloneExpand = false
 
+    /// Clipping windows wrapping each entry snapshot. The morph animates
+    /// these, not the snapshots (see EntryMorphWindowView). Parallel to
+    /// `entrySnapshots` rather than keyed by item index, so repeated indices
+    /// each keep their own window instead of one silently replacing another.
+    private var entryMorphWindows: [(window: EntryMorphWindowView, itemIndex: Int, sourceFrame: CGRect)] = []
+
     /// CADisplayLink-driven paging state. Replaces UIViewPropertyAnimator
     /// because UIVPA's contentOffset interpolation doesn't trigger
     /// per-frame layout invalidation, which breaks parallax and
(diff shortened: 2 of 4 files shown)


bhargav95:
Reported as #221.

bhargav95:
Hey @l984-451, this one is ready too! I've confirmed it works in the simulator. I've added before/after videos to confirm the change. :)

l984-451:
thanks for testing and providing the videos! makes my job a lot easier.

PR #222, opened 2026-07-20
docs: add build, test, and setup notes to CLAUDE.md

bhargav95:
Documentation only — no code changes. Fills in the gaps a fresh clone runs into.

### What's added

**The `Secrets.swift` step.** `RivuletApp.swift` reads `Secrets.sentryDSN`, so the app target won't compile until the gitignored file exists. The template is already there; the doc now says to copy it, and clarifies that `Secrets.swift` is the *only* gitignored file in `Rivulet/Config/` — `TMDBConfig`, `InsightsConfig` and `InputConfig` are tracked and arrive with the clone.

**How to run tests.** The doc had build commands but no test commands. Adds the full-suite invocation, the `-only-testing:` form for a single class and a single method, and a pointer to where tests live (`RivuletTests/Unit/`, with fixtures/helpers/mocks alongside).

**A few facts worth stating once:**

- All three targets use buildable-folder references (`PBXFileSystemSynchronizedRootGroup`), so new files are picked up from disk and never need adding to the `.xcodeproj`.
- No SwiftLint or SwiftFormat is configured, so there is no lint command to run.
- There is no CI test run. `codemagic.yaml` builds, archives and ships to TestFlight on tag push; the two GitHub workflows are the Claude review and mention bots. Tests are local-only.
- `Docs/` and `.claude/` are gitignored, so `Docs/RIVULET_PLAYER.md`, `Docs/DESIGN_GUIDE.md` and the `aether-update` / `rivulet-tvos-uikit` skills referenced throughout CLAUDE.md aren't in the repo. That surprised me on a fresh checkout, so the doc now says so rather than leaving a contributor hunting for missing files.
- Restates the README's point that the Simulator doesn't fully mimic Apple TV, next to the Simulator commands where it's most likely to be read.

**Structure tree:** adds `Services/Insights/` and the in-player `Insights*` panels, which were missing.

### Notes

Every claim was checked against the repo rather than assumed — the pbxproj for the folder references, `.gitignore` for what's actually ignored, `codemagic.yaml` and `.github/workflows/` for CI, and `RivuletApp.swift` for the `Secrets` dependency. The single-method example names a real test (`MediaItemTests/test_id_returnsRef`).

Two judgement calls to reject freely: the `Docs/`-is-gitignored note documents a deliberate choice of yours (`c89b52d`) and may be more detail than you want in a public file, and the Simulator caveat duplicates the README.

Split out from #220 so the fix and the docs stay independently reviewable.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -33,6 +33,7 @@ Rivulet/
 │   │       └── Subtitles/    # SubtitleManager, SubtitleParser, SubtitleOverlayView, SubtitleClockSyncController
 │   ├── LiveTV/         # PlexLiveTVProvider, IPTVProvider, LiveTVDataStore
 │   ├── IPTV/           # M3UParser, XMLTVParser, DispatcharrService
+│   ├── Insights/       # InsightsTriviaClient, InsightsShowIDResolver (in-player cast/trivia panel)
 │   ├── Cache/          # CacheManager, ImageCacheManager
 │   └── Focus/          # FocusMemory (tvOS section focus restoration)
 ├── Views/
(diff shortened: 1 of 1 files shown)


bhargav95:
@l984-451 I have a couple more PRs in review; this is the first. Thank you! 🙏

PR #223, opened 2026-07-20
Settings: one title size for every row kind

bhargav95:
Settings rows use two different title sizes depending on what kind of row they are. This makes them all one size.

## The problem

| Row kind | Title size |
|---|---|
| Navigation, info (drill-in rows) | 36pt |
| Toggle, cycle, action (in-place controls) | 32pt |

A settings page mixes both kinds freely, so the title size steps by 4pt every few rows with no rule the eye can pick up. On an Apple TV it reads as rows being misaligned, not as a hierarchy.

## Screenshots
Before: "Display size" setting font is larger than the other Toggle settings.
After: All settings are the same size
| Before | After |
|:--:|:--:|
| before | after |

## The change

**Commit 1** — one title size for every row kind, and the per-kind `titleSize` parameter is removed so there's no longer a knob that can reintroduce the split.

36pt is the one kept, because navigation rows are the majority on most pages, and the 36pt title / 32pt value pairing that results is what `navigationValue` rows already shipped.

**Commit 2** — title and value labels move from regular to medium weight. The tvOS scale is medium at nearly every level, and light strokes lose definition over TV compression at distance.

The two are split so either can be dropped on its own.

Nothing else changes: no row heights, colors, focus behavior, or layout.

## References

- [tvOS typography](https://developer.apple.com/design/human-interface-guidelines/tvos/visual-design/typography) — the type scale and weights
- [Lists and tables](https://developer.apple.com/design/human-interface-guidelines/lists-and-tables) — disclosure indicators as the navigation affordance

### DIFF
--- a/Rivulet/Views/Settings/UIKit/SettingsCell.swift
+++ b/Rivulet/Views/Settings/UIKit/SettingsCell.swift
@@ -58,7 +58,7 @@ final class SettingsCell: UICollectionViewCell {
         selectedBackgroundView = nil
         contentView.backgroundColor = .clear
 
-        titleLabel.font = .systemFont(ofSize: 32, weight: .regular)
+        titleLabel.font = .systemFont(ofSize: 36, weight: .regular)
         titleLabel.textColor = .white
         titleLabel.translatesAutoresizingMaskIntoConstraints = false
         bg.addSubview(titleLabel)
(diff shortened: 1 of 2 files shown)


bhargav95:
@l984-451 this is ready for review :)

l984-451:
thanks!

PR #225, opened 2026-07-21
Guide fade + Aether and FfmpegBuild version bump

tresby:
- Bump Aether to 5.17.3
- Bump FffmpegBuild to 2.2.0

- Guide background/art now has a fade effect to smooth the transition when navigating the guide.


tresby:
Aether changes breaks Plex Live TV again, ill fix and re-commit

PR #226, opened 2026-07-21
Live TV/Guide updates + Aether/Ffmpeg bump

tresby:
- Added a fade effect to the background/art for the Live TV guide so make navigation appear smoother.

- Added 'Guide Mini-Player' option in 'Live TV' settings (opt in). Replaces the top right artwork with a minimised player to let the user navigate the guide while still enjoying playback.

- Changed EPG navigation to prevent 'Back/Menu' from exiting guide - now will always focus back on the currently playing program (if mini-player active).

- Updated Aether code to let Plex Live TV continue to function correctly.

- Bump AetherEngine version to 5.17.3

- Bump FfmpegBuild version to 2.2.0

- Adjusted Live TV subtitle settings (custom renderer) to mimic 'slide' effect behind the OSD when the OSD is bought up.

### DIFF
--- a/Rivulet.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Rivulet.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -6,8 +6,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/superuser404notfound/AetherEngine",
       "state" : {
-        "revision" : "b05d384705792e949f22b0bd7c98ac8453cbd208",
-        "version" : "5.8.4"
+        "revision" : "54b0b864ca1669865596cd3b348be367a23f1ba9",
+        "version" : "5.17.3"
       }
     },
     {
@@ -15,8 +15,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/superuser404notfound/FFmpegBuild",
       "state" : {
-        "revision" : "e8ae797a8ac9b715dfb6805b4929cc9f3e4173f2",
-        "version" : "2.1.0"
+        "revision" : "e832f22c89909297b7abe3a94c05b2816414b054",
+        "version" : "2.2.0"
       }
     },
     {
--- a/Rivulet/Services/Plex/Playback/AetherPlayer.swift
+++ b/Rivulet/Services/Plex/Playback/AetherPlayer.swift
@@ -482,19 +482,14 @@ final class AetherPlayer: PlayerProtocol {
         )
     }
 
-    /// Live TV load. Sets `isLive` so the engine treats the source as live
-    /// (seek becomes a no-op, live-edge/reconnect behavior). HLS sources use
-    /// `nativeRemoteHLS` so AVPlayer plays the remote playlist directly (no
-    /// demuxer probe / loopback — "HLS straight to AVPlayer"); everything else
-    /// (raw MPEG-TS, etc.) goes through the engine's demux/remux to a loopback
-    /// HLS stream. Either way the engine publishes `currentAVPlayer` for the
-    /// AVKit OSD. No start position (live).
-    /// `forceEngineDemux` disables the native-HLS shortcut so the engine's own
-    /// demuxer opens the playlist instead — used as a fallback when AVPlayer's
-    /// native path fails against a server (e.g. a Plex transcode session that
-    /// rejects AVPlayer's request pattern).
-    func loadLive(url: URL, headers: [String: String]?, forceEngineDemux: Bool = false) async throws {
-        let isHLS = Self.isHLSURL(url) && !forceEngineDemux
+    /// Live TV load. Plain HLS uses Aether's native remote-HLS route. Plex HLS
+    /// needs `hlsIngest`: its MPEG-TS segments carry broadcast audio and
+    /// DVB/teletext subtitle streams that AVPlayer's remote-HLS route would not
+    /// expose. Aether 5.17+ requires those playlists to enter through
+    /// `HLSLiveIngestReader`; handing a playlist to the raw URL reader now fails
+    /// closed instead of relying on FFmpeg network demuxing.
+    func loadLive(url: URL, headers: [String: String]?, hlsIngest: Bool = false) async throws {
+        let nativeRemoteHLS = Self.isHLSURL(url) && !hlsIngest
         let options = LoadOptions(
             suppressDisplayCriteria: false,
             httpHeaders: headers ?? [:],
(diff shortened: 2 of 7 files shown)


tresby:
@l984-451 
Hey mate,

Couple of changes here - let me know what you think of them.

Struggled bit with navigation around the mini-player. Finding it hard to let 'back' scroll back to the current program, but then open the sidemenu once press again without navigating. Change any navigation you don't find intuitive should you want to add the feature.

Also on the blurred background - I originally wanted to mimic the old plex guide but had found it hard to recreate the colour gradient/mesh/blur effect they got - so settled with a larger imaged blurred. If you can re-create it like the following:
Plex-Live-TV-Breaking-Bad-1536x864

that is what my original aim was.

As always, hopefully not too much of a pain to untangle my vibe code!

Thanks!

l984-451:
this is Claude, not Bain:

Reviewed this against current `main`. I built it merged with main (Package.resolved conflict resolved to main's 5.18.2 pin): **BUILD SUCCEEDED**, no new warnings. So this is not a compile problem. Two of the four things here are good, one is arguably fixing a live bug on main, but the packaging is broken and the mini-player has a stuck state and a Menu trap.

Requesting changes.

## Blockers

**1. Drop the Package.resolved commit (`3399f27`).**

- It sets AetherEngine 5.17.3. `main` is already at 5.18.2 (`db37bdf`), so this is a downgrade.
- GitHub reports the PR mergeable but that is stale. A real merge conflicts on Package.resolved; the branch is 6 commits behind main.
- The branch's own `project.pbxproj` still pins `exactVersion 5.8.4`. SPM honours the exact pin, not a hand edited lockfile, so **the PR as authored cannot compile**: `HLSLiveIngestReader(playlistURL:httpHeaders:)` does not exist in 5.8.4 (only `public convenience init(playlistURL:)`; the headers init is internal there).

Rebase on main and drop the lockfile commit. The engine bump is already done.

**2. The player can never be stopped, and Menu can never leave the guide.**

With the mini player enabled: guide Menu goes to `requestPlayingProgramFocus`, and fullscreen Menu with chrome closed hits the `dismiss` override which calls `onMinimize` instead of dismissing. `onDismiss`, the only thing that clears `activePlayerController` and lets the session die, is unreachable in normal use; it is only nil'd on stream resolve failure or channel replacement. So once a channel starts there is no in-guide way to stop the audio, and Menu never backs out of Live TV. That is a tvOS convention violation rather than a preference. Better shape: Menu moves focus to the playing programme if it is not already there, and otherwise falls through to normal back.

Related, needs device verification: with `.sidebarAdaptable` the tab content is often kept alive. Please leave Live TV via the sidebar with a mini player running and confirm audio actually stops.

**3. `presentPlayerController` has a silent dead end.**

It retries 12 x 20ms and then just returns. If it gives up, `displayMode` is stuck at `.fullscreen`, where `guideContent` is `.opacity(0).disabled(true)`, with nothing presented and `guideMenuHandler` nil. Result: black screen, no focus, no Menu handler, audio still playing. It needs a failure path back to `.minimised`.

## The Aether routing change: right instinct, too wide a blast radius

The underlying problem is real. 5.18.2 carries AE#140, which makes the raw live path fail closed when the body starts with `#EXTM3U` (`AetherEngineError.hlsPlaylistOnRawLivePath`). 5.8.4 had no such guard. Main's `forceEngineDemux: url.path.hasPrefix("/livetv/sessions/")` hands exactly that kind of extension-less playlist to the raw reader, so Plex tuned-session Live TV is plausibly broken on main right now. That is a consequence of the 5.18.2 bump on our side, not something this PR caused.

But `shouldUsePlexHLSIngest` is "Plex && looks like HLS", which also captures `/video/:/transcode/universal/start.m3u8`. That URL works today on `nativeRemoteHLS`. And because `loadLive` gates the software-decode force on `!nativeRemoteHLS`, every Plex live channel now also gets `setForceSoftwarePathForTesting(true)`, losing hardware decode on the transcode leg.

Please narrow it: ingest for the extension-less session Part key, keep `.m3u8` on `nativeRemoteHLS`, and decouple the interlace software-path force from the ingest decision. If the motivation for routing start.m3u8 through ingest was mp2 or teletext surviving a direct-stream remux, that is a defensible argument, but it needs the actual failing URL plus the engine error in the PR rather than an assumption.

Two smaller notes on this area:

- `isHLSDelivery` matches `/livetv/.../tune`, which is a POST returning XML and never reaches the player. Dead condition.
- The sibling hole is untouched: `MultiStreamViewModel` loads slots through `AetherPlayer.load(url:...isLive:true)`, which never sets `nativeRemoteHLS`, so multiview on a Plex HLS channel hits the same AE#140 rejection. Out of scope for this PR, but worth knowing.

## Lower severity

- `requestPlayingProgramFocus` sets `selectedGroup = nil`, silently wiping the user's category tab on every Menu press.
- Selecting a non-airing cell while minimised is a silent no-op, so Select looks dead.
- `cv.remembersLastFocusedIndexPath = false` is only restored in `didUpdateFocus` when the pending target lands. If it never lands, the guide loses focus memory for the rest of the session.
- The backdrop fade chains 0.18 + 0.016 + 0.55 + 0.42s of `Task.sleep`, so the crisp artwork effectively never appears while browsing, and the 16ms "commit a frame" sleep bakes in a frame-rate assumption. A single `.id(url)` plus `.transition(.opacity)` crossfade gets most of this for a fraction of the machinery. No image quality regression though; `CachedAsyncImage` was already thumb quality.
- No changelog entry. Every user-facing change and every AetherEngine bump needs a line in `WhatsNewView.changelogs`.

## Fit

- **Subtitle chrome scrim** (Live TV captions sliding behind the OSD, matching VOD): small, consistent, keep.
- **HLS ingest fix**: needed, keep, narrowed. Worth splitting into its own PR so it can land immediately, since main is likely broken for Plex tuned sessions today.
- **Backdrop fade**: fine in principle, over-engineered in practice.
- **Mini player**: good concept, and it correctly deletes the already-dead `.pip` / `LiveTVPlayerView` guide path. The implementation reparents a `UIViewController` between a SwiftUI representable and a modal presentation, held in `@State`, with retry loops, `DispatchQueue.main.async` hops and a `preservePlaybackForReparenting` flag. That is exactly the SwiftUI-hosting-UIKit fragility the project has been migrating away from. If it ships, it should own the player in UIKit.

## UI/UX changes in this PR

- Guide backdrop crossfades between programmes instead of cutting.
- New Settings row, Live TV > Guide Mini-Player (toggle, default off) with a descriptor. Follows the title-only plus left-panel-description rule correctly.
- When enabled, the top-right guide artwork is replaced by live video.
- Menu in the guide no longer exits Live TV while a mini player is active.
- Live TV subtitles now lift and slide behind the OSD like VOD.
- `.pip` display mode removed (it was unreachable).

tresby:
Will seperate the major mini-player stuff and have another go
Then will do a smaller guide fade commit because that one should be a simpler one
Aether got bumped in the mean time, apologies

PR #237, opened 2026-07-22
Update GuideLayoutView.swift

tresby:
Guide ambient background and 16x9 art fade effect commit only. Makes navigating guide feel a lot smoother and scrolling less jarring.

### DIFF
--- a/Rivulet/Views/LiveTV/GuideLayoutView.swift
+++ b/Rivulet/Views/LiveTV/GuideLayoutView.swift
@@ -118,6 +118,18 @@ struct GuideLayoutView: View {
     @State private var focusedChannel: UnifiedChannel?
     @State private var focusedProgram: UnifiedProgram?
 
+    // Backdrop transition state. Keep the existing artwork treatment, but let
+    // the wash settle before revealing the crisp image for the new programme.
+    @State private var outgoingBackdropImage: UIImage?
+    @State private var incomingBackdropImage: UIImage?
+    @State private var backdropProgress: Double = 1
+    @State private var displayedArtworkImage: UIImage?
+    @State private var artworkOpacity: Double = 0
+
+    private let artworkFadeOutDuration = 0.18
+    private let backdropFadeDuration = 0.55
+    private let artworkFadeInDuration = 0.42
+
     private let tick = Timer.publish(every: 30, on: .main, in: .common).autoconnect()
 
     var body: some View {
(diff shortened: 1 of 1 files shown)


Rivulet/Views/LiveTV/GuideLayoutView.swift
```diff
@@ -118,6 +118,18 @@ struct GuideLayoutView: View {
...
+    // the wash settle before revealing the crisp image for the new programme.
+    @State private var outgoingBackdropImage: UIImage?
+    @State private var incomingBackdropImage: UIImage?
+    @State private var backdropProgress: Double = 1
+    @State private var displayedArtworkImage: UIImage?
+    @State private var artworkOpacity: Double = 0
+
+    private let artworkFadeOutDuration = 0.18
+    private let backdropFadeDuration = 0.55
+    private let artworkFadeInDuration = 0.42
```
bhargav95, line 131: I tested in the simulator. This is quite slow feeling as it's all in sequence: fade out artwork -> fade in backdrop -> fade in artwork, summing up to 1.15 sec!

I tried 0.12, 0.30, and 0.25s respectively, it's better, but the sequentially fade-ins are still distracting.

CC @l984-451 

```suggestion
    private let artworkFadeOutDuration = 0.12
    private let backdropFadeDuration = 0.30
    private let artworkFadeInDuration = 0.25
```

bhargav95: @tresby I have a PR here for overlapping the fades. You can additionally shorten the durations if required.
- https://github.com/tresby/Rivulet/pull/1

tresby: The duel step was intentional to re-create the old effect from the native apple plex guide.

Although I do like the duel effect - happy to leave it up to Bain which he prefers - either fade will make guide navigation less jarring in my opinion.

bhargav95: Fair enough! If we want the fade-ins to be sequential, we could lower the duration. The PR works otherwise. ✅

tresby: Yeah happy to go with whatever really - see what the boss thinks ;)

l984-451:
Merged, thanks @tresby. Follow-up on main in `a65b6b8`.

To settle the fade question you two were debating: you were both right about your own half, and the thing making it feel wrong was neither.

`transitionBackdrop` mutated state before its first `await`. `.task(id:)` cancels on every focus change, so `artworkOpacity = 0` landed, the sleep threw, and everything after it never ran. At a normal browsing pace — moving faster than the crossfade — the artwork faded out and never came back. The full sequence only completed if you stopped and waited it out, which is why it read as slow rather than broken. Shortening the durations narrows that window but doesn't close it.

So the fix was ordering, not timing: the settle window now runs before any mutation, so a cancel leaves the backdrop untouched, and it doubles as a debounce (holding a direction across the guide shouldn't fire an image load per channel).

What survived from the PR: your fade-out-before-swap ordering, which is the part that stops the old artwork flashing against the new wash, and the `blurredBackdrop` / `crispArtwork` split, which was worth doing on its own. @bhargav95's overlap went in too — running wash and artwork together is both quicker and cheaper, since it halves how long two full-screen blurs stay composited. Also swapped the 16ms `Task.sleep` for `Transaction(disablesAnimations:)`, which does the same job without racing a frame boundary.

Durations are 180ms settle, 0.35s fade. Those are a judgement call, not a measurement — still to be checked on an Apple TV, since the Simulator isn't honest about either guide scrolling or blur compositing.

PR #241, opened 2026-07-23
feat: configurable skip length in Playback settings

### LINKED ISSUE #233: Skip Length Options
### Before submitting

- [x] I have searched for existing feature requests

### Describe the feature

Support for other skip length options besides 30 seconds. 5-30, increments of 5?

### Use case

I find that 30 seconds is a bit too much for my use case, and prefer 10-15 typically. Not a deal breaker, but would be a ‘nice to have’ option.

### Any additional context

_No response_

bhargav95:
## Summary

Adds a **Skip Length** setting (Settings → Playback) so a single Left/Right press can skip **5, 10, 15, or 30 seconds** instead of a fixed 30. Default stays 30, so existing behavior is unchanged. The chosen value drives the tap skip everywhere it is used today: the VOD player, the Control Center skip buttons, and the Live TV multiview jump.

Closes #233.

## Details

- New `SkipInterval` enum + picker page, following the existing Autoplay Countdown settings pattern (no new plumbing).
- `InputConfig.tapSeekSeconds` now reads the `skipSeconds` preference (default 30) instead of a hardcoded constant.
- Options are limited to magnitudes that ship numbered `goforward`/`gobackward` SF Symbols (5/10/15/30), so the on-screen seek indicator always shows the number.

## Testing

- New unit tests cover the option set, the unset default (30), the stored round-trip, and an end-to-end check that a tap skip dispatches the configured amount.
- Build and tests pass on the tvOS Simulator.

## Demo

https://github.com/user-attachments/assets/4d06a895-3e74-43d0-82c3-30e7b97b855d

Screenshot 2026-07-22 at 5 14 05 PM
Screenshot 2026-07-22 at 5 14 15 PM

### DIFF
--- a/Rivulet/Config/InputConfig.swift
+++ b/Rivulet/Config/InputConfig.swift
@@ -30,8 +30,10 @@ enum InputConfig {
 
     /// Single-press Left/Right skip. Applied uniformly across every remote and
     /// focus state (content-focused, IR d-pad, keyboard, and the focused
-    /// scrubber). May become a user-facing Playback setting later.
-    static let tapSeekSeconds: TimeInterval = 30
+    /// scrubber). User-configurable via Settings → Playback → Skip Length.
+    static var tapSeekSeconds: TimeInterval {
+        TimeInterval(SettingsStore.int(SkipInterval.storageKey, default: SkipInterval.defaultValue.rawValue))
+    }
     static let jumpSeekSeconds: TimeInterval = 30
 
     static let dpadThreshold: Float = 0.3
--- a/Rivulet/Services/Plex/Playback/NowPlayingService.swift
+++ b/Rivulet/Services/Plex/Playback/NowPlayingService.swift
@@ -32,6 +32,7 @@ final class NowPlayingService: ObservableObject {
     private var cachedArtwork: MPMediaItemArtwork?
     private var cachedArtworkURL: String?
     private var interruptionObserver: NSObjectProtocol?
+    private var skipIntervalObserver: NSObjectProtocol?
     private var inputCoordinator: PlaybackInputCoordinator?
     private var lastHandledPlaybackState: UniversalPlaybackState?
     /// Track if we've set Now Playing info with valid duration (> 0)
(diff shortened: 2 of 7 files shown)


bhargav95:
@l984-451 Let me know if this enhancement is okay to do! I've been able to verify it in the simulator ✅

l984-451:
### Code review

Found 3 issues:

1. `SettingDescriptor` no longer has an `iconColor` field — the branch predates `b64f0de` ("drop unused iconColor from SettingDescriptor"), so the new `"skipLength"` entry will not compile against current `main`. This hunk merges *cleanly*, so a naive rebase produces a non-building tree.

https://github.com/l984-451/Rivulet/blob/af82bfe48c6ad2f2f8bc20e976761c38824a922c/Rivulet/Views/Settings/SettingsDescriptors.swift#L154-L159

2. `pageInfo(for:)` uses bare `.blue` rather than `.systemBlue`. This region *does* produce real conflict markers against `ad197e2` ("settings page colors are UIColor"), and resolving it by keeping the PR side hard-fails: `UIColor` has no `.pink` (only `.systemPink`), which the pre-refactor `case .music` still uses. Correct resolution is to keep main's system-color switch and add `case .skipIntervalPicker: return ("forward.fill", .systemBlue)`.

https://github.com/l984-451/Rivulet/blob/af82bfe48c6ad2f2f8bc20e976761c38824a922c/Rivulet/Views/Settings/SettingsDescriptors.swift#L430-L432

3. The Control Center skip buttons do not pick up the setting after it changes. `preferredIntervals` reads `InputConfig.tapSeekSeconds` inside `setupRemoteCommandCenter()`, which runs exactly once from the `NowPlayingService` singleton's `init`. That was safe while `tapSeekSeconds` was a `static let`; now that it is user-editable, the interval freezes at whatever was set when the singleton was first created, and the handler dispatches `skipEvent.interval` verbatim. In-app tap-seek updates live, Control Center stays stale until relaunch — so the PR description's claim that the value drives "the Control Center skip buttons" only holds until the user changes it.

https://github.com/l984-451/Rivulet/blob/afb829ff233a2b8e482c27f91810796ad7a07361/Rivulet/Services/Plex/Playback/NowPlayingService.swift#L277-L291

bhargav95:
@l984-451 Fixed!

l984-451 (approved):
Re-reviewed at 39951ef. The three issues from the earlier review are resolved: the descriptor matches the current SettingDescriptor shape, the picker page uses .systemBlue, and the Control Center skip commands now seek by the live InputConfig.tapSeekSeconds with the displayed glyph refreshed per playback session. Approach and scope look right. Thanks for the follow-up.

PR #250, opened 2026-07-25
Release trickplay thumbnails on item end; merge live TV channels off main

bhargav95:
Two independent fixes, one commit each.

## 1. Trickplay thumbnails were never released

**Impact:** unbounded memory growth across a viewing session. Each BIF holds
every trickplay frame of an item (~270 JPEGs for a 45-minute episode), and
`clearCache(partId:)` / `clearAllCache()` had **no callers anywhere** — so a
15-episode session retained all 15. On tvOS that is a jetsam risk, not just
footprint.

Tracks the preloaded part and releases it in two places:

- `stopPlayback()` — normal teardown.
- `playNextEpisode()` — the one that matters. It swaps `metadata` in place and
  never calls `stopPlayback()`, so teardown alone would miss the binge case
  entirely. Sits beside the existing `insightsRecheckTask` cancel, which drops
  outgoing-item state for the same reason.

A later scrub re-fetches on demand.

## 2. Live TV channel merge ran on the main thread

**Impact:** UI freeze two seconds after launch for IPTV users with large
playlists — the merge and sort of every channel blocked the thread that drives
the focus engine, while the home screen was still loading.

`LiveTVDataStore` is `@MainActor`, and a `Task {}` created inside a `@MainActor`
method inherits that isolation, so the sort ran on main. With a large M3U that
is ~1.7M comparisons, falling through to Unicode string collation because most
playlists omit `tvg-chno`. `Task(priority: .background)` does not help —
priority does not change isolation.

The tell: the `await MainActor.run { self.channels = ... }` right after the sort
was a no-op, implying the body was assumed to be off-main.

`loadEPG` in the same file already solves this. This copies that pattern —
snapshot `providers`, use `Task.detached` — and the existing `MainActor.run`
becomes the single real main hop. Network fetches were already concurrent and
are unchanged.

## Testing

`xcodebuild build` and `xcodebuild test` both succeed (full suite);
`swiftlint lint --strict` reports 0 violations. No new warnings in either file.

**Not exercised on device.** Change 2 in particular is worth checking on real
hardware with a large playlist — the Simulator does not reflect tvOS
main-thread behaviour well.

## Review notes

- No changelog entry — did not want to guess the next build key in
  `WhatsNewView.swift`. #1 is arguably user-facing.
- #1 bounds growth per session but adds no LRU cap in `PlexThumbnailService`.
  Happy to put a hard ceiling there instead if you prefer.

### DIFF
--- a/Rivulet/Services/LiveTV/LiveTVDataStore.swift
+++ b/Rivulet/Services/LiveTV/LiveTVDataStore.swift
@@ -319,15 +319,22 @@ class LiveTVDataStore: ObservableObject {
     private func updateSourceInfo() async {
         var infos: [LiveTVSourceInfo] = []
 
+        // One pass instead of a filter per provider: on a large lineup the
+        // per-provider filter is O(providers × channels) on the main actor,
+        // a second main-thread sweep right behind the channel load.
+        var channelCounts: [String: Int] = [:]
+        for channel in channels {
+            channelCounts[channel.sourceId, default: 0] += 1
+        }
+
         for (id, provider) in providers {
             let isConnected = await provider.isConnected
-            let channelCount = channels.filter { $0.sourceId == id }.count
 
             infos.append(LiveTVSourceInfo(
                 id: id,
                 sourceType: provider.sourceType,
                 displayName: provider.displayName,
-                channelCount: channelCount,
+                channelCount: channelCounts[id] ?? 0,
                 isConnected: isConnected,
                 lastSync: nil  // TODO: Track last sync time
             ))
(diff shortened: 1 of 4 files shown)


l984-451 (changes requested):
Thanks for this — both diagnoses are correct, and the reasoning in the PR body made them fast to verify. Two things needed before merge, one per commit. Neither is in code you wrote; both are in the functions this PR starts calling, which is what makes them worth catching now.

**1. `9800f03` — the release is unreliable while a BIF is still loading**

Confirmed the leak is real: `clearCache(partId:)` and `clearAllCache()` had zero call sites anywhere in `Rivulet/` or `RivuletTests/`. Placement is right too — `playEpisode(_:)` delegates to `playNextEpisode()`, so the one insertion covers both transition paths, and releasing via the stored `preloadedThumbnailPartId` rather than `metadata` correctly sidesteps the metadata swap earlier in the function.

The problem is in `PlexThumbnailService.clearCache(partId:)`. It nils `bifCache[partId]` and forgets `unavailableParts`, but leaves `loadingTasks[partId]` alone, and `preloadBIF` writes back asynchronously:

```swift
let task = Task<BIFData?, Never> { await loadBIF(...) }
loadingTasks[partId] = task
Task {
    if let result = await task.value { bifCache[partId] = result }   // lands after the clear
    else { unavailableParts.insert(partId) }
    loadingTasks[partId] = nil
}
```

Release a part while its BIF is downloading and that trailing task repopulates the cache afterwards. `preloadedThumbnailPartId` has already advanced, so nothing will ever release it — the exact leak this PR fixes, made permanent for that part. It triggers on advancing episodes faster than a BIF download, which is ordinary binge behaviour and the case the commit is aimed at.

This is latent today because nothing calls `clearCache`; this PR is the first caller, so it activates it. The fix belongs in `PlexThumbnailService`. Worth noting `loadingTasks[partId]?.cancel()` alone isn't enough — the trailing task captured `task` directly and still performs the write. It needs the write guarded (check `loadingTasks[partId] === task` before assigning) or an epoch counter.

Minor, take it or leave it: `unavailableParts.remove(partId)` inside `clearCache` discards the "this part 404s" knowledge, so a later scrub retries a known-failing fetch. Pre-existing and self-correcting.

**2. `999e059` — `refreshChannels()` has the same bug and is untouched**

The isolation analysis is right, and the dead `await MainActor.run` is good evidence for it. I checked the Swift 6 side since the detached task captures `self` and snapshots the provider dictionary: `protocol LiveTVProvider: Sendable`, and `@MainActor` classes are implicitly Sendable, so both crossings are legal. No strict-concurrency exposure.

The gap is `refreshChannels()`. It runs the identical task group and `allChannels.sort` with no `Task` wrapper at all — directly in the `@MainActor` body, so it is unambiguously sorting on main. That's the path that runs when a user adds or edits an IPTV source, which is exactly when a large M3U is in play. Fixing `loadChannels` alone means the freeze still reproduces through a different entry point. It isn't copy-paste either: `refreshChannels` assigns `self.channels` directly with no `MainActor.run`, so it needs the snapshot-and-detach written out rather than lifted.

Two cosmetic notes: the new comment says the `MainActor.run` is "the only main hop", but `await self.updateSourceInfo()` on the next line is one too; and `Task.detached` drops priority inheritance, which is almost certainly irrelevant here but is a real difference from `Task {}`.

**Also worth knowing**, out of scope for this PR: `chapterThumbnails` in `UniversalPlayerViewModel` is per-item state keyed by chapter index and never cleared on episode swap. Beyond the bounded leak, if the new episode has a chapter at an index the old one populated and its own fetch is partial, it can render the previous episode's thumbnail. Same bug class you were already hunting — happy to see it as a follow-up, please don't grow this PR for it.

On verification: both changes are worth a device pass rather than the Simulator. For the BIF commit specifically, one behaviour change to sanity-check is that re-entering an item and immediately scrubbing now shows no thumbnail until the BIF re-downloads. Fine trade against jetsam, but worth confirming it feels acceptable.

Requesting changes on that basis. Good PR — the split into two commits with the reasoning written out made this a lot easier to review.

l984-451:
Correcting the record: I posted that review against a stale head and didn't re-check before submitting.

`80201f2` had already fixed both points, and went further than I asked — the same race in `getThumbnail` (which I missed entirely), keeping `unavailableParts` rather than clearing it (you were right there and I was wrong), and extending the off-main fix to `loadEPG`.

GitHub won't let me dismiss a review on a closed PR, so this comment is the correction. The work is good and main needed it: landed as `4cd9d32` with your authorship. Sorry for the noise.

Issue #192, opened 2026-07-02
Back button on remote takes you all the way back to Home


grigsby:
### Before submitting

- [x] I have searched for existing feature requests

### Describe the feature

Pressing the “back” button on the remote should eventually take you back to the Home tab.

### Use case

After browsing a library, pressing the back button the remote only takes you back to opening the left sidebar, but still on the library tab. One more press of the back button should return back to the “Home” tab. (This is how the official Plex app works. Pressing the back button enough times will eventually return you to the Home tab. Interestingly, it seems like in the Apple TV and Apple Music apps, “back” eventually exits the app, which I think is even less useful. So I guess this is just a personal preference.)

VMolly:
Good catch! I noticed that too and if possible would be great for the back button to consistently take you back to the home screen with single continuous presses. I am used to pressing the back button to exit the app but that is just a personal preference

PR #251, opened 2026-07-26
Staged Menu back: return the page to its top row (#19)

### LINKED ISSUE #19: Request: Back button should scroll back up to top row
If you scroll down on any page, the only way to get back to the top is to scroll all the way back up. I admit, I’m used to the Plex behavior where if I press the back button, the selection resets all the way back to the first item in the first row of that tab.

bhargav95:
Closes #19.

## The issue

Scroll down a few rows on Home (or a library) and the only way back to the top is to scroll all the way up. Pressing Menu jumps straight out to the sidebar instead, so you lose your place. Plex resets to the first item of the first row first, which is what #19 asks for.

This was previously investigated and recorded as platform-blocked. That conclusion was wrong, though its individual findings were all correct — see below.

## The solution

`UIWindow.sendEvent(_:)` receives the press before gesture recognizers and before the responder chain. Returning early there withholds it from the system entirely.

`MenuPressInterceptor` swizzles that method — the same technique already used for the sidebar focus guard in `TVSidebarView` — and offers the press to registered surfaces. Each surface decides for itself whether it currently owns focus, so anything with a player, detail page or Settings on top declines and Menu keeps its normal meaning there.

`PlexHomeViewController` conforms to `MenuBackHandling`: below the top row it consumes the press and returns the page to its top row; at the top row it declines and the press passes through untouched.

## Screencap 
Go down a few rows, click Menu once to return to hero, click Menu again to open sidebar

https://github.com/user-attachments/assets/68b4515c-d438-4fd1-ab1f-7097dc15344e

## Behavior

| Where focus is | Menu does |
|---|---|
| Below the top row on Home | Returns to the top row, focus on the hero's Play. Sidebar stays shut. |
| Below the top row in a library | Same — the surface is mode-agnostic |
| Below the top row on Discover | Same |
| Below the top row on Search (no hero) | Returns to the first section |
| On the hero / top row | Passes through; the system reveals the sidebar |
| In the sidebar, not on Home | Existing behavior: selects Home (#192) |
| In the sidebar, on Home | Unchanged |
| Detail page open | Declines; the detail dismisses as before |
| Player, tile menu, Settings | Declines; unchanged |

So from deep in a page it takes three presses to walk out: top row, then sidebar, then Home.

## Testing

- 12 unit tests covering the stage-1 policy and the press-phase swallow state machine (pairing the withheld `.began` with its `.ended` so the system never sees half a press)
- Full suite passes; SwiftLint clean
- Verified in the tvOS 26.5 simulator against a real Plex library: deep on Home, deep in a library, at the top row, and with a detail page open

## Why it looked impossible

While focus is in the content area, a Menu press never reaches the responder chain. Measured on tvOS 26.5, none of these ever fire for it:

- `pressesBegan` on the content view controller
- `pressesBegan` on the `UIWindow`
- `shouldUpdateFocus(in:)` on the content collection view
- SwiftUI `.onExitCommand`
- any menu-press `UIGestureRecognizer` — a census of the whole hierarchy finds exactly one, and it belongs to `.onExitCommand`, not the sidebar

The `.sidebarAdaptable` sidebar reveal happens above all of that. Every seam checked previously sits *below* the one layer that does see the press first.

## Not covered

- Search mode is implemented but was not exercised in the simulator
- Simulator only so far — this is focus and remote-input behavior, which the simulator does not fully mimic, so it wants a run on an Apple TV before merging
- The separate launch-focus flash (sidebar briefly focused before the hero) is pre-existing on main and untouched here
- The `onExitCommand(perform: nil)` change that would restore exit-to-Home-Screen at the top level is deliberately left out — it is a separate behavior change from this issue

### DIFF
--- a/Docs/superpowers/specs/2026-07-10-menu-back-navigation-design.md
+++ b/Docs/superpowers/specs/2026-07-10-menu-back-navigation-design.md
@@ -3,9 +3,26 @@
 Date: 2026-07-10
 Issues: #19 (Back button should scroll back up to top row), #192 (Back button on remote takes you all the way back to Home)
 
-## OUTCOME (2026-07-10) — read this first
+## SUPERSEDED (2026-07-27) — read this first
 
-**Shipped: #192 only. #19 abandoned as platform-blocked.**
+**#19 is no longer platform-blocked. Both stages ship.** Stage 1 is
+`MenuPressInterceptor`, which takes the press at `UIWindow.sendEvent(_:)` and offers it to
+the focused surface; `PlexHomeViewController` consumes it below the top row and declines at
+the top, so the system still performs its native sidebar reveal. Stage 3 is unchanged from
+what shipped for #192.
+
+The 2026-07-10 conclusion below was right about every seam it tested and wrong about the
+one it did not test. `sendEvent(_:)` sits earlier in the event pipeline than `pressesBegan`:
+the window sees the press before the `.sidebarAdaptable` shell consumes it, so returning
+early there withholds it from the system entirely. The dead-seam list is still accurate and
+still worth reading before anyone proposes a fourth approach — it is the reason the
+interceptor sits where it does rather than anywhere more obvious.
+
+Note the asymmetry that made this easy to miss: a `pressesBegan` swizzle **on `UIWindow`**
+was tested and does not see `.menu`, so the window looked ruled out as a class of hook. It
+was not; only that method on it was.
+
+## OUTCOME (2026-07-10) — historical, superseded above
 
 The staged design below assumed the app could intercept the Menu press that moves focus
 from a content grid into the sidebar. It cannot. Empirically confirmed on the tvOS 26
--- a/Rivulet/Views/Components/WhatsNewView.swift
+++ b/Rivulet/Views/Components/WhatsNewView.swift
@@ -141,6 +141,7 @@ struct WhatsNewView: View {
 
     static let changelogs: [(version: String, features: [String])] = [
         ("1.0.4 (70)", [ // confirm build number at release
+            "Pressing Menu now takes you back to the top of the page first, so you no longer lose your place jumping straight out to the sidebar",
             "Opening a movie or show now grows its poster proportionally instead of stretching it out of shape",
             "Settings rows now all use one title size and weight, so each page reads evenly from top to bottom",
             "Updated AetherEngine to 5.23.3. Fixes titles that never finished playing at the very end, resuming near the end of a movie getting stuck on Loading, Dolby Vision being dropped when you resume close to the end, and a crash after several minutes of playback on fast connections. Also: some MP4 movies that opened to a frozen frame now play and use less power, Dolby Vision titles streamed from a slow server no longer jump the scrubber backward or stall when you seek, the Atmos badge is now accurate, long Live TV sessions no longer slowly use up memory until the app closes, and leaving a Dolby Atmos title no longer leaves your receiver looping the last sound",
--- a/Rivulet/Views/Media/PlexHome/UIKit/PlexHomeViewController.swift
+++ b/Rivulet/Views/Media/PlexHome/UIKit/PlexHomeViewController.swift
@@ -4669,7 +4669,11 @@ extension PlexHomeViewController: UICollectionViewDelegate {
         // The staged Menu back has landed; stop routing focus to the top. A
         // directional move clears it too — if the top cell never took focus the
         // flag would otherwise latch and yank focus back on some later,
-        // unrelated update. Mirrors the two exits `needsInitialHeroFocus` has.
+        // unrelated update. `needsInitialHeroFocus` additionally requires the
+        // move to start INSIDE the collection, because it is armed at launch
+        // when focus can arrive from the sidebar. This flag is only ever armed
+        // by handleMenuBack(), which already requires focus inside the
+        // collection, so that conjunct would be dead weight here.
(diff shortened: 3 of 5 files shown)


l984-451:
### Code review

Found 2 issues:

1. `returnToTopRow()` arms `needsInitialHeroFocus` without the `heroState` guard its consumer requires, so focus can strand on the hero's loading placeholder. The consumer only clears the flag once the hero has content (`let heroHasContent = kind == .hero && heroState != .loading`), and the comment there explains why: "The loading placeholder parks focus but must not consume the initial-hero routing." The producer re-arms unconditionally on `heroSectionIndex != nil`. Scroll deep on a library page while the hero hub is still resolving, press Menu, and focus parks on the placeholder with the flag still armed until content arrives. Mirroring the consumer's `heroState != .loading` check on the producer side closes it.

https://github.com/l984-451/Rivulet/blob/228ec655e0f44506ce5737a2c04411d36fb46ef1/Rivulet/Views/Media/PlexHome/UIKit/PlexHomeViewController.swift#L4835-L4843

consumer-side guard for comparison:

https://github.com/l984-451/Rivulet/blob/228ec655e0f44506ce5737a2c04411d36fb46ef1/Rivulet/Views/Media/PlexHome/UIKit/PlexHomeViewController.swift#L4683-L4690

2. `MenuPressSwallowState.shouldWithhold` does not check `isSwallowing` before re-entering the `began` branch, so a second `.began` arriving before the first `.ended` re-invokes `handle()` and overwrites the pending press's tracking. `returnToTopRow()` is idempotent so the visible impact is small, but the state machine's whole purpose is pairing a withheld `.began` with its matching terminal phase, and this is the one path that breaks the pairing. `testHandlerIsAskedOnlyOnBegan` covers intermediate phases of a single press, not a second overlapping press. A `guard !isSwallowing` on the `began` branch fixes it.

https://github.com/l984-451/Rivulet/blob/228ec655e0f44506ce5737a2c04411d36fb46ef1/Rivulet/Views/TVNavigation/MenuPressInterceptor.swift#L33-L43

Notes, not blocking:

- Swizzling `UIWindow.sendEvent(_:)` is a materially broader seam than the `shouldUpdateFocus(in:)` precedent in `TVSidebarView` that the doc comment cites as prior art. That one declines a focus proposal; this one withholds raw press events process-wide for the lifetime of the process. The technique looks correctly implemented (original IMP captured before replacement, concrete-subclass targeting, one-time-per-class guard, main-actor confinement), and the guards in `handleMenuBack()` mean a covered page declines. Calling out the difference in scope, not objecting to it.
- The `wantsTopFocus` clear omits the `prevInside` conjunct that `needsInitialHeroFocus` uses, though the accompanying comment says it "mirrors" it. Unreachable today, since `wantsTopFocus` is only armed via `handleMenuBack()`, which already requires focus inside the collection view. Worth correcting the comment so the asymmetry is not mistaken for an oversight later.
- Simulator-only verification is disclosed in the PR body, which is the right call per CLAUDE.md. The two issues above are both in paths the unit tests do not reach, so a device pass is worth doing after they are addressed.

bhargav95:
@l984-451  Fixed one, Claude code pushed back on the other 😅 

> [!NOTE]
> The reply below was written by Claude Code and posted from my account — it is not my own commentary.

> Thanks — one of the two is real, the other I'd push back on.
>
> **2. Overlapping `.began` — agreed, fixed in 1e905c0.** A second `.began` before the first press ended re-invoked `handle()` and overwrote the pending press's tracking, which is precisely the pairing the state machine exists to keep. Added `!isSwallowing` to the `began` branch, plus two tests: one for the overlapping press, and one asserting the *declined* path stays re-askable (a decline leaves no swallow pending, so the next `.began` must be a fresh decision — easy to break with a naive guard).
>
> **1. `needsInitialHeroFocus` without the `heroState` guard — I don't think this is a defect.**
>
> The mechanism you describe is accurate, but parking on the placeholder is the intended behavior rather than a strand. `HeroLoadingCell` is deliberately focusable:
>
> ```swift
> // Quiet focus parking spot — no visible focus treatment, no actions.
> override var canBecomeFocused: Bool { true }
> ```
>
> So in that scenario focus lands on the placeholder, and because the flag stays armed, focus advances to Play the moment content arrives — identical to the launch path, which is the behavior `returnToTopRow()` is deliberately reusing. The flag remaining armed is what makes that second hop happen; it's the design, and the consumer-side comment you quoted is describing exactly this contract from the other end.
>
> Mirroring `heroState != .loading` onto the producer would regress it. The `else` branch routes through `wantsTopFocus`, which clears as soon as focus lands on the top section. When the hero then gains content the diffable identity changes (`hero-loading` → `hero-overlay`), the cell is replaced, and with no flag armed focus is not re-asserted — it's lost to whatever the engine picks next. That's strictly worse than parking on the placeholder and then advancing to Play.
>
> **On the notes:**
>
> - `wantsTopFocus` / `prevInside` — fair, the comment overclaimed. Corrected in the same commit: `needsInitialHeroFocus` needs that conjunct because it is armed at launch when focus can arrive from the sidebar; `wantsTopFocus` is only ever armed by `handleMenuBack()`, which already requires focus inside the collection, so the conjunct would be dead weight. Now stated explicitly so the asymmetry doesn't read as an oversight.
> - Swizzle scope — agreed, and worth having on the record. `sendEvent` is a much broader seam than the `shouldUpdateFocus` precedent. It's the only layer that sees the press at all while focus is in the content area (the other five are enumerated in the PR body), so the breadth is forced rather than chosen, but the comparison to prior art in the doc comment does undersell it.
> - Device pass — agreed, still outstanding, and noted in the PR body as a merge blocker.

PR #252, opened 2026-07-27
Below-fold info cards

bhargav95:
## What

Renders the below-fold Information / Languages / Accessibility trio as cards matching the About section, and makes them openable.

Previously these were three bare columns on a darkened band, sitting directly under two proper About cards — the same kind of supporting metadata in two unrelated treatments.

- Information / Languages / Accessibility are now cards with the About card's fill, radius, and focus lift. The darkened band is gone.
- Each is its own focus stop, and Select opens that section in a popup. The card keeps trimming long genre and studio lists to stay scannable; the popup carries the full list.
- The episode detail page shares `InfoColumnsView`, so it gets the cards too (non-focusable there, since it scrolls normally).
- `AboutCardControl` becomes `DetailCardControl`, owning its own content inset so both surfaces stop restating the padding.

Two things fixed along the way:

- **The Related shelf header was silently dropped whenever Related landed first** (a movie with no trailers or extras — the common case). It was suppressed on the primary peek row on purpose, but that means the section loses its label exactly when it's the only thing there. Now always shown.
- **Short info popups scrolled by 8pt and clipped their last line.** The scroll view's top inset was applied once at init and only the bottom one was ever reset, so content that fit the card exactly still sat 8pt low. Both insets now track the overflow state, and the sizing is a pure `InfoPopupCardLayout` with tests.

## Verified

- `xcodebuild build` for tvOS Simulator
- `swiftlint lint --strict` — 0 violations
- Full unit suite passes; 7 new popup-layout tests, each written failing first
- Driven in the Simulator: cards render at equal height with no clipping, Left/Right moves across all three, Select opens the matching section, the Related header appears on a movie with no extras, and the synopsis popup no longer clips

## Not verified

**Focus behaviour has only been seen in the Simulator**, which per `CLAUDE.md` can't be trusted for the focus engine. The three focus stops and the moves between them need a device pass.

## Known follow-ups

- The info section's height is a hardcoded `504` that encodes the cards' internal padding in a file with no other knowledge of them. A title with many audio/subtitle languages could overflow it.
- The card-Select popup and the hero Info button now render different bodies for the same section — `fullInfo` has Region of Origin, container progress, and codec-rich audio; the card has Genres and Studio. Reconciling them is a product decision, so it's deliberately left alone.

### DIFF
--- a/Rivulet/Views/Media/MediaDetail/UIKit/BelowFoldCollectionView.swift
+++ b/Rivulet/Views/Media/MediaDetail/UIKit/BelowFoldCollectionView.swift
@@ -244,8 +244,12 @@ final class BelowFoldCollectionView: UIView, UICollectionViewDelegate {
             case .related:
                 // Home-identical shelf: one full-width ShelfRowCell hosting
                 // the posters (margin 40, symmetric peeks, pitch landings).
+                // As the primary row, lift the section by its own header (+ the
+                // section's 8pt pad) so the 40pt peek strip is POSTER, like the
+                // episodes row — the "Related" title rides just above it.
+                let relatedLift = ShelfRowCell.headerHeight + 8
                 return self.homeShelfHostSection(
-                    topInset: isPrimary ? peek : 0,
+                    topInset: isPrimary ? max(0, peek - relatedLift) : 0,
                     sectionBottomInset: isPrimary ? 36 : 56
                 )
             case .cast:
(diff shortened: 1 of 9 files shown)


l984-451:
@bhargav95 I know this is a draft, but figured I'd opine on it anyways. I need to recheck the ATV app, but I don't think it has this behavior and not sure its needed. The below fold area needs some love and cleanup, but I don't think it should be clickable as a popup - it should just contain all the information there on screen. 

The other fixes in here are great, thank you.

bhargav95:
@l984-451 I checked the ATV before working on this actually, and tried to replicate it here 😅 . I'd like to have it as cards but I agree it doesn't have to open up in a popup.
Yeah, it's still in draft as the diff blew up, I want to see if I can scope it better or split it into smaller PRs.

l984-451:
haha well then nevermind! I stand corrected. Keep going, thanks!

PR #254, opened 2026-07-27
Live tv refresh fix + OSD channel selection

tresby:
# Live TV: keep the guide fresh, and add a channel list to the player

Two related Live TV changes: the guide now refreshes itself when it has gone stale, and the player OSD gains a channel list so you can change channels without going back to the guide.

## 1. Guide refresh

**Problem.** Coming back to Live TV often showed an empty or out-of-date guide. Two causes:

- Every Live TV surface only loads when its data is **empty** (`if dataStore.epg.isEmpty { … }`), so once populated, nothing ever re-fetched it for the life of the process.
- The EPG window is anchored at load time. After the app has been suspended for a few hours, every programme in the loaded window is in the past, so the grid renders blank even though `epg` is non-empty. This is the "it just isn't loaded" case.

**Change.** `LiveTVDataStore` now records when channels and EPG last loaded successfully, and exposes a single entry point:

```swift
func isStale(maxAge: TimeInterval = defaultMaxAge, now: Date = Date()) -> Bool
func refreshIfStale(maxAge: TimeInterval = defaultMaxAge) async
```

Stale means **either** the grid is older than 30 minutes **or** `epgLoadedThrough` no longer reaches the current time. Age alone would miss the second case (the channels tab only loads a 6 hour window), which is the one that produces a blank guide.

Called from:

- `UIApplication.willEnterForegroundNotification`, observed inside the store so it works regardless of which screen is showing
- `LiveTVContainerView.task` and `ChannelListView.task`

Notes on the implementation:

- **Concurrent callers share one refresh.** The tab appearing and the foreground notification routinely fire together, and two overlapping `loadEPG` calls cancel each other through `epgLoadTask`, leaving the grid empty — the exact failure this is meant to fix.
- **A failed load does not count as fresh.** If every provider throws, the timestamps stay unset so the next visit retries rather than sitting on an empty list for 30 minutes.
- No new settings, no new user-facing controls.

## 2. Channel list on the player OSD

The Live TV rail reuses the Up Next button slot (`onUpNext`, relabelled with a TV icon) to open a channel list, mirroring the VOD OSD's season list.

`ChannelListPanelView` is deliberately built to the same visual contract as `UpNextListView`: same 520pt panel width, 92pt rows, 120x68 16:9 thumbnail, corner radii, focus scale, accent bar on the current row, and the same "open scrolled to the current row" behaviour including the `prepareForPresentation` layout timing that the Up Next panel documents.

Content is live rather than episodic:

- Channel name, and the current programme beneath it
- 16:9 programme art from `UnifiedProgram.landscapeURL`. Anything else is aspect-**fit** rather than filled: `iconURL` / `posterURL` are routinely 2:3 posters, and cropping a portrait poster to 16:9 leaves only a band across its middle. Channel logo is preferred over a poster as the fallback, since identifying the channel matters more than a sliver of programme art.
- A small progress bar across the bottom of the artwork showing how far through the current programme is, matching the resume bar on the VOD play pill. The track is darkened rather than translucent white, because here it sits on artwork rather than on a dark button.

Two separate row types rather than one generic list: the row data, artwork source, and image lifetimes have nothing in common between an episode and a channel. The shared look is held by matching the metrics.

**Channel switching is in place.** The first version dismissed the player and presented a new one for the chosen channel. That failed on device: this VC is presented from a SwiftUI host whose view leaves the window hierarchy while the player covers it, so the re-present landed on a detached controller (`Attempt to present … whose view is not in the window hierarchy`). The new player loaded its stream but never appeared, and the outgoing one kept running because its teardown is driven by `viewDidDisappear`, which never came.

So the playback lifecycle is now split into `startPlayback()` and `teardownPlaybackSession()`, with dismissal and channel switching both going through the same teardown so the two cannot drift. Teardown covers the stream task, Plex keepalive, join telemetry, legible output, the engine player and its subscriptions, and it resets the fallback ladder and clears the outgoing channel's subtitle cues.

The channel button also sits at the front of the cluster (left of Subtitles) and is the rail's default focus target on Live TV, via a new `defaultFocusButton` that leaves VOD's Subtitles-first behaviour untouched. Changing channels is the primary live action, so it gets the first position and the first highlight.

## Fit

- No new settings and no new user-facing configuration. The refresh threshold is a constant.
- The channel list reuses the existing rail button, panel presenter, and row styling rather than introducing a new surface.
- Both changes are UIKit on a UIKit surface; no SwiftUI added to a primary surface, no `UIHostingController` in a cell, no UIKit context menus.
- Arguably the guide refresh is something the app should always have done, rather than a feature.

## Testing

Built and exercised on device for the channel list, the in-place switch, and the button placement.

**Not yet verified:**

- The 30 minute staleness path end to end (needs a long idle, or a temporary shorter threshold).
- The foreground refresh fires even when the user never opens Live TV. It no-ops without configured sources, but it is a network fetch on every foreground while stale. Happy to gate it on "Live TV has been visited this session" if that is preferred.
- Behaviour with a source whose line-up changes between refreshes.

## Open questions for review

1. **Fallback art order.** Channel logo is preferred over programme poster when there is no 16:9 art. If most sources carry posters and few logos, the other order may look better.
2. **Progress bar is a snapshot** taken when the panel opens rather than ticking live. Programmes run for tens of minutes, so it cannot visibly drift in the seconds the panel is on screen.
3. **Join latency on switch.** Switching channels pays a full join, same as opening from the guide. Back-to-back switching feels it. AetherEngine's `liveJoinProfile: .fastZap` (5.19.0) would address that, but it is a separate change with its own trade-off.


tresby:
Built as a replacement to the mini-player. Think this is a little nicer and fits in with the design as it appears like the episode selector on VOD.

l984-451:
@tresby I just merged some PRs and it created conflicts for this. Can you fix them? thanks!

tresby:
@l984-451 think its just the 'what's new' that's the conflict now - unsure if you wanted to change that to suit some of your other changes or I can delete it from the PR?

PR #256, opened 2026-07-28
fix: keep the Related header and peek posters

bhargav95:
Split out of the below-fold info-cards work so it can be reviewed on its own.

Two fixes to the detail page's Related row:

- **The header was dropped whenever Related landed first.** The section
  passed `header: !isPrimary`, so a movie with no episodes — where Related
  is the primary row — rendered the shelf with no "Related" title at all.
  The header is now always drawn.
- **The resting peek showed a clipped title instead of posters.** As the
  primary row the section is lifted by its own header height (+ the 8pt
  section pad), so the 40pt peek strip is poster, matching the episodes
  row, with the title riding just above it.

`ShelfRowCell.headerHeight` becomes internal so the layout can size the
lift from the same constant the cell uses, rather than a second literal.

Verified with a tvOS Simulator build and `swiftlint --strict` (0
violations). The peek geometry itself is a focus/layout change, so it is
worth a look on a real Apple TV before merge.


l984-451 (approved):
Reviewed: no issues found.

PR #257, opened 2026-07-28
fix: stop short info popups from scrolling and clipping

bhargav95:
Split out of the below-fold info-cards work so it can be reviewed on its own.

The info popup sized its card to the measured content, but the scroll
view's `contentInset` was set once at construction (`top: 8, bottom: 72`)
and only the bottom inset was ever cleared. So a popup whose content fit
the card exactly still carried the 8pt top inset: the content was pushed
down inside a card sized without it, the card became scrollable by 8pt,
and the last line was clipped.

The card height and both insets now come from one value, `InfoPopupCardLayout`,
with a single writer applying it. The insets are breathing room for content
scrolling in and out of the card's edges, so they belong only to the
overflowing case — non-scrolling popups get neither, and their offset is
reset.

`InfoPopupCardLayout` is a pure struct, so the sizing rules are covered by
unit tests rather than needing a device: short content hugs and doesn't
scroll, content exactly at the cap doesn't scroll, tall content caps at the
screen fraction and keeps its breathing room, and the forced-height path
always scrolls.

7/7 tests pass on the tvOS Simulator; `swiftlint --strict` reports 0
violations.

### DIFF
--- a/Rivulet/Views/Media/MediaDetail/UIKit/InfoPopupViewController.swift
+++ b/Rivulet/Views/Media/MediaDetail/UIKit/InfoPopupViewController.swift
@@ -17,6 +17,45 @@
 
 import UIKit
 
+/// Card height + scroll insets for a measured content height. Pure so the sizing
+/// can be tested without a device (see InfoPopupCardLayoutTests).
+///
+/// The insets are breathing room for content scrolling in and out of the card's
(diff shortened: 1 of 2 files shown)


l984-451 (approved):
Reviewed: no issues found.

PR #258, opened 2026-07-28
Match Apple's caption rendering in the subtitle overlay

tresby:
Size, box and placement now follow AVPlayer's own caption rendering instead of fixed point values.

Size is a fraction of the presentation height times the system's relative-character-size multiplier, so it tracks Subtitles & Captioning across the whole range. The multiplier previously did not: the clamp in CaptionAppearance was 0.5...2.0, and tvOS reports 0.35 at the smallest setting, so the bottom of the range was silently flattened.

The box hugs the text more tightly, and its padding and corner radius scale with the point size rather than staying fixed, so a large caption size no longer draws big glyphs in a small hard-edged box.

The resting bottom margin is measured from the picture rather than the screen, so letterboxed content is captioned on the image instead of in its black bar. AetherPlayer publishes the item's presentation size for this; when it is unknown (the software path has no AVPlayer) the overlay falls back to its full bounds, which is the previous behaviour. With the rail up, the screen-anchored rail clearance still wins.

This should be pretty much an exact mimic visually and placement wise for the Native AVPlayer sub renderer - while keeping additional options like delay and height adjustment toggles.

### DIFF
--- a/Rivulet/Services/Plex/Playback/AetherPlayer.swift
+++ b/Rivulet/Services/Plex/Playback/AetherPlayer.swift
@@ -57,6 +57,23 @@ final class AetherPlayer: PlayerProtocol {
     /// (embedded or sidecar) is selected and the engine has cue data.
     @Published private(set) var isSubtitleActive: Bool = false
 
+    /// The item's presentation size (pixel dimensions after aperture/pixel
+    /// aspect correction), for hosts that must place UI against the VIDEO
+    /// rect rather than the screen — the subtitle overlay measures its bottom
+    /// margin from the picture, so captions on 2.39:1 content sit above the
+    /// letterbox rather than down in the black bar.
+    ///
+    /// `.zero` until an item reports one, and on the software path (no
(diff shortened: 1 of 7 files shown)


tresby:
Bigger subs changes probably coming with Aethers latest integration @l984-451 - some better styling support for text based. 

Once that's nutted, I can have a go at ASS subs.

I'll keep working on the latest changes, but I'll wait to submit once Aether is bumped and if you commit these changes if you're happy. Don't want to submit a PR too broad or with too much overlap. :)

tresby:
With Aether changes I've implemented a bit of an overhaul of the subs.

These changes made our subs look almost identical to the system subs and scaled them a lot better with the settings.

Upcoming PR I will do tonight:

Same base but included all system settings + override settings.

Have also made the custom renderer the default for all media (SW, Native HLS, VOD remux).

Bumped Aether and FFmpeg to support styling in cues (Position, colour, font (not ass yet), size etc)

Will note it all in the PR.

Although it will be a larger PR, hopefully since its all just a sub renderer update its not too bad to review @l984-451

PR #259, opened 2026-07-29
Text-based subtitle overhaul/refinement

tresby:
# Match Apple's caption rendering, and keep content styling and placement

Subtitles are rendered by Rivulet's own overlay rather than by AVKit, and had
drifted from what tvOS itself draws — wrong size at the ends of the system size
range, a background box per line, a margin measured from the screen, and every
piece of styling the content specified discarded except colour.

This makes the overlay match AVPlayer's rendering, surfaces the styling and
placement AetherEngine now parses, and gives Live TV's native-HLS path the same
renderer as everything else.

---

### Caption rendering matches tvOS

- **Size** tracks Subtitles & Captioning across its whole range. Point size is a
  fraction of the presentation height times the system's relative-size
  multiplier, matching how Apple derives it, rather than a fixed size.
- **Background** is one tight box per cue sized to the longest line, not a box
  per line. Padding and corner radius scale with the type, as Apple's do — a
  fixed 4pt radius reads as a hard rectangle at large caption sizes.
- **Resting margin** is measured from the **picture**, not the screen, so a
  2.39:1 film is captioned on the image instead of down in the black bar.

### Content styling and placement now render

AetherEngine 5.26.0 parses the full ASS override set out of the event line
libavcodec produces for *every* text format, so bold, italic, underline,
strikethrough, colour, font face and size arrive for SRT, WebVTT, ASS and
broadcast teletext through one path — as does the position the source asked
for, which is how broadcasters move captions to the top of frame to avoid
covering on-screen graphics.

Each attribute is gated independently by the matching system **Video Override**
state, read per-attribute from MediaAccessibility, so content wins only where
the user allows it to. Position has no system setting, so a content position is
honoured subject only to the floor below.

### One placement floor, every cue, both players

Captions keep the same margin (6% of picture height based off AVPlayer) off whatever they rest
above: the bottom of the picture with the rail hidden, the top of the rail with
it showing. A cue carrying its own placement keeps that placement while it sits
at or above the floor, and is raised to the floor otherwise — so a WebVTT
`line:100%` or a bottom teletext row lands where captions belong rather than
against the edge of the picture, while a cue the broadcaster put at the *top* of
frame still goes there.

Both overlays take the floor from `SubtitleAdjustments`, so captions do not
shift when a title falls back from the engine route to the HLS one. The lift
animates on the chrome's own curve and slides up *behind* the rail.

### Live TV's remote HLS path (due to sub changes)

On `nativeRemoteHLS` the engine never demuxes, so the stream's WebVTT cues
belong to AVPlayer. They are now pulled out through an
`AVPlayerItemLegibleOutput` (`suppressesPlayerRendering = true`,
`textStylingResolution = .sourceAndRulesOnly`) and drawn by the same overlay as
every other path. One renderer is authoritative everywhere, and the user's
caption settings apply on that path for the first time.

### Height stepper is sticky per media

It was global. It is now keyed per title and per channel like Delay, defaults to
0, and applies to the default caption band only — a cue carrying its own
placement is exempt, since an authored position must not drift with an app-level
offset.

---

## Why the engine bump is here

It is a dependency, not a drive-by. The styled-run fields land in **5.26.0**,
cue placement only reaches a host from **5.28.1** (until then every cue was
stripped of it on insert), and standalone `.vtt` sidecars need the webvtt
demuxer that arrives with **FFmpegBuild 2.3.0**. None of this works on 5.23.11.

FFmpegBuild moves 2.2.0 → 2.3.0 transitively. It still builds FFmpeg n8.1.2 and
dav1d 1.5.1, so that licence entry is unchanged, and the shipped frameworks
still declare `MinimumOSVersion 26.0`, matching `TVOS_DEPLOYMENT_TARGET`. The
LGPL written offer in `OpenSourceLicenses` names the exact linked tag and commit,
so it moves with the pin.

---

## Notes for review

Things that look wrong but aren't, or are easy to break later:

- **`CaptionAppearance` reads each `allowsContent*` immediately after its own
  MediaAccessibility call.** Every read overwrites the shared `behavior`
  out-param, so reordering these silently attributes one setting's override
  state to another.
- **The font-scale clamp widened from `0.5...2.0` to `0.25...4.0`.** tvOS
  reports 0.35 at the smallest setting, so the old bound silently flattened the
  ends of the system's own range. This was the actual cause of captions not
  tracking the size setting; the size fraction is tuned against it and should
  not be re-tuned to compensate for a clamp problem.
- **Live TV route detection keys off the routing decision, confirmed against the
  item's URL — not off `subtitleTracks` being empty.** AE#154 makes the engine
  mirror AVPlayer's legible options into `subtitleTracks`, so the old emptiness
  test now reports the opposite of what it used to. The URL check also follows
  an engine-side #168 reroute onto the loopback, which the routing decision
  alone would miss.
- **The Live TV overlay's hosting controller sets `safeAreaRegions = []`.**
  Every margin in that overlay is measured from the screen; a hosting controller
  otherwise applies tvOS's title-safe inset to its content, which made the rail
  gap roughly double its intended size. VOD gets the same effect from
  `.ignoresSafeArea()`.
- **The rail-driven overlay rebuild wraps its `rootView` assignment in
  `withAnimation`.** That assignment is a UIKit-side write, not a SwiftUI
  transaction, so the overlay's `.animation(value:)` has nothing to attach to
  and captions jump rather than slide. VOD needs none of this — there
  `controlsVisible` comes off an `@Published`.
- **Overlay z-order is now explicit in Live TV.** It was correct but incidental,
  resting purely on `mountSubtitleOverlay()` being called before `setupChrome()`.
  Stated at both ends now, matching the precedent `PlayerContainerViewController`
  already sets.
- **`AetherPlayer.videoSize` observes `presentationSize` in two hops** rather
  than through a `\.currentItem?.presentationSize` key path — nested KVO through
  an optional chain is not reliably delivered, and a missed update would leave
  the overlay measuring against a stale aspect ratio for the session.
- **The height stepper's value is pushed into the overlays by the host** rather
  than read there with a dynamic-key `@AppStorage`. The key changes under a
  channel switch or next-episode swap while the view keeps its identity, and
  `@AppStorage` re-binding across that is not worth relying on.
- **Bitmap cues now map against the video rect**, not the full bounds.
  `SubtitleImage.position` is normalised against the source frame, so on
  letterboxed content PGS/DVB cues were being stretched vertically and pushed
  toward the black bar.

---

## Testing

Device and Simulator (tvOS 26.5), against:

- DVB-T broadcast teletext (AU FTA, page 801) on the software path — coloured,
  positioned, multi-row cues
- HLS fMP4/HEVC channels carrying WebVTT renditions on the native path
- VOD with embedded, sidecar and bitmap (PGS) subtitles
- Letterboxed 2.39:1 content, for the picture-relative margin
- Rail show/hide and scrubbing, for the lift

Checked on device rather than Simulator where the behaviour is chrome- or
focus-adjacent, per the repo's note that the Simulator does not fully mimic
Apple TV.

---

## Known gap, engine-side

Multi-row teletext cues keep their column padding at interior row boundaries —
trailing spaces at the end of row 1, leading spaces at the start of row 2. The
engine trims the outer edges of a cue and folds blank rows, but nothing trims at
a retained newline, so a two-line caption renders slightly off-centre. Reported
upstream with a suggested fix; it needs no change here.


tresby:
@l984-451 This pretty much completes text-based subs IMO. Hopefully you're okay with this all being 1 PR.

If this is committed - I can then work on ASS compatibility and then all subs should be sorted by Aether and our own sub renderer, instead of some cases being Native AVPlayer subs. Hopefully easier to maintain going forward.

tresby:
May need to wind back Live HLS using custom renderer. Further testing on an Apple TV surfaced some errors.

Will resubmit when ready

PR #281, opened 2026-08-03
Fix guide group focus navigation and top padding

tresby:
- Adjusted top padding due to sidebar pill obstructing program poster.

- Guide groups were still in swift - now are uikit.

- Fixed not being able to navigate back and forth between the guide and the guide group bar.


tresby:
Anything you need me to change on this one @l984-451 ?
Can’t use the TestFlight build as I can’t access the EPG without this fix atm.
Sorry if it’s just a time constraint! Appreciate all the other commits! App is coming along very nicely I think!

l984-451:
@tresby No worries! yeah, just balancing work and real life and this lol. Working on merging it.

I'll be more helpful on the live tv front soon since american football is starting back up and I'll start using live tv more often.

tresby:
Thanks @l984-451

Yes completely understand!

Yeah that would be great! Saw some feedback on the plex forum regarding Plext based Live TV from another Aus user and they're having trouble with channels loading.

Would love to pick your brains about what direction we should go...

Essentially my findings are:

1. We don't tell plex enough codecs we actually support 
2. Because plex delivers Live TV via HLS - we try to play via AVPlayer directly essentially - even when we should (Aus TV is interlaced so should go through SW)

So I feel like we should ask Plex the scanType while tuning and base our path (for plex tunes) on that (Aether probe can't detect scanty reliable in h264)? Is there much OTA tv that's interlaced in America or is it a niche issue?

Sort of have a build doing what I've explained there so if you agree with that logic I can finish it off and submit.

PR #300, opened 2026-08-12
Dvb_teletext safezone

tresby:
Teletext wasn't obeying same safe zone as webvtt was

### DIFF
--- a/Rivulet/Views/Player/UIKit/CaptionOverlayView.swift
+++ b/Rivulet/Views/Player/UIKit/CaptionOverlayView.swift
@@ -212,11 +212,18 @@ final class CaptionOverlayView: UIView {
         /// quantises the 24-row grid to three bands and supplies no percentage.
         ///
         /// Free to tune, because nothing measurable pins it: the source
-        /// genuinely does not say where in the top third the caption belongs.
+        /// genuinely does not say where in the third the caption belongs.
         /// Set near where the same page's proxied WebVTT resolves (it carries an
         /// exact `line:`, typically around 10%), so the two routes look similar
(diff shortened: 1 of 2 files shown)


l984-451:
### Code review

Found 1 issue:

1. The bottom-band change breaks `testCoarseLeftBandUsesTheColumnAnchorAndFloor`. That test uses `alignment: 1, position: nil`, which is `row == 0`, so it hits the rewritten `else` branch. On its 1000x1000 canvas the picture fills the view, so `letterbox` is 0 and `floor` is 6% = 60. The new `max(band, lowest)` picks `band` = 100 and puts the box's `maxY` at 900, but the test asserts 940. CLAUDE.md notes there is no CI test run, so this will not be caught before merge.

https://github.com/l984-451/Rivulet/blob/68dc943650d33d5646fc62917c44aee73a973c01/Rivulet/Views/Player/UIKit/CaptionOverlayView.swift#L473-L478

https://github.com/l984-451/Rivulet/blob/68dc943650d33d5646fc62917c44aee73a973c01/RivuletTests/Unit/Playback/CaptionOverlayPlacementTests.swift#L74-L80

tresby:
Will fix and push the fix.

Another note on this - I have these safezone for webvtt and teletext - but may remove the safezone from webvtt. 

The reason being teletext cues from aether arent very specific, where as webvtt usually are. I think i originally implemented it for webvtt due to subs being placed from the middle of the sub, not say, the top left corner like avplayer.

Will fix that too and submit sperately.

tresby:
fixed + other change sin new PR

PR #301, opened 2026-08-12
Plex Live TV: direct-play the tuned session, and group the guide

tresby:
Attempt to make Plex Live TV more robust

We now request a few more compatible codecs from plex to reduce server-side remuxing and look at what plex tells us its sending back to decide on how we play content back.

Also added some grouping support for multiple tuners and for plex favourites

Let me know if there's anything that doesn't fit @l984-451 

Claude Summary:

# Plex Live TV: direct-play the tuned session, and group the guide

Plex grants direct play on a tuned live session and Rivulet was throwing the
grant away — falling back to AVPlayer, which cannot decode the mp2 audio or DVB
teletext that direct play exists to preserve. This makes the grant reachable,
and while in the same code, fixes a URL-encoding bug that had been corrupting
the client profile on one of four build sites.

Also groups the guide by tuner and adds a Favourites tab, which is why the diff
spans both playback and the guide.

---

### The client profile declares what the engine can actually decode

Audio gains `dts, truehd, flac, alac, opus, vorbis, pcm`, and both direct-play
clauses gain `subtitleCodec=dvb_teletext`.

Video is deliberately **unchanged** at `h264, hevc, mpeg2video`. Broadcast does
not carry anything else, and declaring codecs the container never holds buys
nothing.

The subtitle clause is the one that matters most. The profile previously
declared no subtitle codec for direct play while telling PMS to convert
subtitles to WebVTT, so a stream matching on video and audio could still be
pulled into a conversion purely by its captions — and the teletext the caption
renderer wants would be gone.

### A granted direct play is an HLS playlist, not raw MPEG-TS

The part key PMS returns is
`/livetv/sessions/{uuid}/{consumer}/index.m3u8`. `forceEngineDemux` routed that
to the engine's **raw** live path, which fails closed on a playlist body:

```
[AVIOReader] HLS playlist body on the raw live path (AE#140); failing closed.
             Use LoadOptions.nativeRemoteHLS or HLSLiveIngestReader for m3u8 sources.
```

The host's failure ladder then re-tuned and landed on `nativeRemoteHLS` — so
every granted direct play was quietly played by AVPlayer, with no teletext and
no mp2. It also cost **two tuner grabs** per channel: one to fail, one to fall
back.

`HLSLiveIngestReader` demuxes the playlist's underlying transport stream, so the
cues reach `SubtitleModel` → `CaptionOverlayView` like every other engine path.
If the reader cannot open the playlist (encryption, fMP4) it retries natively on
the **same** session rather than throwing, so the previous behaviour is still the
floor and no second tuner is grabbed.

### The profile's clause separators were being destroyed on rebuild

`X-Plex-Client-Profile-Extra` joins its clauses with `+`, which is a legal query
character and therefore not percent-encoded by `URLComponents` — so it reaches
PMS as a space and the clause boundary is lost. It has to travel as `%2B`.

Three build sites applied that pass. `PlexLiveTVProvider.buildStreamURL` did
not: it re-parses an already-correct URL to swap the session UUID, and reading
`queryItems` decodes `%2B` back to a raw `+`. Visible in a device log as
`…vorbis%252Cpcm)+add-direct-play-profile(…`.

All four now go through one `finalizedLiveURL` helper, so the rule cannot be
missed again.

### Scan type decides the deinterlacing path

`loadLive` forced the software backend on every engine-demux live load, for
deinterlacing. That was written when the engine had no deinterlacer on its
native path; it has had a hardware one since 5.8.4.

PMS reports `scanType` on the video stream in the tune response, so a channel
that is genuinely progressive no longer pays for it. **Absence means
interlaced** — only an explicit `"progressive"` opts out, because a needless
software decode costs CPU while a missed deinterlace is combing the user can
see. It reads the server's scan type rather than the codec's own field order,
which broadcast H.264 is known to mis-signal.

---

### Every DVR, not just the first

`fetchChannelsFromGrid` used `Dvr?.first`. A server with two tuners exposes two
DVRs with separate lineups, so one lineup's channels were silently missing from
the guide entirely. One unreachable DVR is now logged and skipped rather than
failing the whole fetch.

### Tuner groups, named from the lineup

The lineup string already carries a human name in its fragment, so no extra
request is needed:

```
lineup://tv.plex.providers.epg.cloud/5fc76c88…#Freeview - Perth (58 channels)
                                               → "Freeview - Perth"
```

Plex's `(58 channels)` suffix is dropped — it is a count for Plex's own picker
and goes stale the moment a channel is hidden. The DVR key is deliberately not a
fallback: "161" is a worse heading than no heading.

Groups appear **only when there is more than one DVR**. With one tuner the tab
would list exactly what All Channels lists.

### Plex Favourites

Favourites are account-level, not server-level —
`epg.provider.plex.tv/settings/favoriteChannels`, a fourth provider host
alongside the three in the Discover section. That is why nothing under
`/livetv/dvrs/…` carries a favourite flag, and why one list can span a local
tuner and Plex's free channels.

`FavoriteChannel[].id` is the same `channelIdentifier` the grid and tune
endpoints use, so no mapping is needed. Array order is the user's own
arrangement and is preserved as `favouriteRank`.

---

## Notes for review

Things that look wrong but aren't, or are easy to break later:

- **The favourites call uses `PlexAuthManager.shared.authToken`, not
  `self.authToken`.** `PlexLiveTVProvider` is constructed with
  `selectedServerToken`, and the provider hosts reject a server token — the same
  account-vs-server trap the Discover endpoints document. `self.authToken` looks
  right and is not.
- **`X-Plex-Provider-Version: 5.1` is required on that call.** Without it the
  endpoint returns a valid `MediaContainer` with no `FavoriteChannel` array
  rather than an error, which reads as "no favourites".
- **Favourites is not a `groupTitle`.** A favourite keeps its tuner group as
  well, and the tab has to honour the arranged order, neither of which a single
  string can express — hence `isFavourite` / `favouriteRank` on
  `UnifiedChannel`. Both are defaulted, so no existing call site changes and the
  IPTV path is untouched.
- **`needsDeinterlacing` treats an absent scan type as interlaced.** That is
  deliberate; do not "simplify" it to a boolean the other way round.
- **Codec lists travel as `%2C`, clauses as `%2B`.** Both look like mistakes.
  Changing either to a raw character makes PMS parse one codec named
  `"h264,hevc,…"` and refuse direct play on every channel, silently.
- **Channel dedupe across DVRs is first-wins**, so a channel two tuners both
  carry is attributed to whichever DVR the server listed first rather than
  appearing twice under two group tabs.

---

## Testing

Observed on device (tvOS 26.5) against a PMS with two DVRs — one XMLTV, one
cloud-EPG — on an HDHomeRun tuner:

- The tune succeeds and PMS reports `scanType=interlaced` for AU FTA channels
- `/decision` returns **direct play** with the widened profile
- Both DVR lineups are fetched, where previously only one was
- The account favourites endpoint returns the expected three channels in order
- The raw `+` corruption was reproduced in a device log before the fix

### DIFF
--- a/Rivulet/Services/LiveTV/PlexLiveTVProvider.swift
+++ b/Rivulet/Services/LiveTV/PlexLiveTVProvider.swift
@@ -152,12 +152,33 @@ actor PlexLiveTVProvider: LiveTVProvider {
             throw error
         }
 
+        // Favourites are account-level, not per-DVR, so this is ONE call for the
+        // whole lineup. Rank is array position: that is the order the user
+        // arranged, and the merged channel list is sorted by number, which would
+        // otherwise throw it away. Best effort — an account with none, or a call
+        // that fails, simply yields no Favourites tab.
+        //
+        // NOT `self.authToken`. This provider is constructed with
+        // `selectedServerToken`, and a server token is rejected by the provider
+        // hosts — the same account-vs-server trap the Discover endpoints
+        // document. Read the account token directly for this one call.
+        var favouriteRanks: [String: Int] = [:]
+        if let accountToken = await PlexAuthManager.shared.authToken {
+            for (index, identifier) in await networkManager
+                .getFavoriteChannelIdentifiers(authToken: accountToken)
+                .enumerated() {
+                favouriteRanks[identifier] = index
+            }
+        }
+
         // Convert to UnifiedChannel
         let channels = plexChannels.map { plexChannel in
-            plexChannel.toUnifiedChannel(
+            let rank = plexChannel.channelIdentifier.flatMap { favouriteRanks[$0] }
+            return plexChannel.toUnifiedChannel(
                 sourceId: sourceId,
                 serverURL: serverURL,
-                authToken: authToken
+                authToken: authToken,
+                favouriteRank: rank
             )
         }
 
@@ -373,7 +394,10 @@ actor PlexLiveTVProvider: LiveTVProvider {
         }
 
         components.queryItems = queryItems
-        let newURL = components.url ?? originalURL
+        // Reassigning queryItems decoded the profile's %2B separators back to
+        // raw '+'. Without this the rebuilt URL reaches PMS with spaces between
+        // the client-profile clauses. See finalizedLiveURL.
+        let newURL = PlexLiveTVChannel.finalizedLiveURL(components) ?? originalURL
 
         // Log session ID regeneration (GitHub #64 - DVB diagnostics)
         let breadcrumb = Breadcrumb(level: .info, category: "plex_livetv")
@@ -480,6 +504,9 @@ actor PlexLiveTVProvider: LiveTVProvider {
                     if let ratingKey = tune.ratingKey {
                         items.append(URLQueryItem(name: "rivuletLiveRatingKey", value: ratingKey))
                     }
+                    if let scan = tune.videoScanType {
+                        items.append(URLQueryItem(name: "rivuletLiveScanType", value: scan))
+                    }
                     items.append(URLQueryItem(name: "X-Plex-Token", value: authToken))
                     dp.queryItems = items
                     if let url = dp.url { return url }
@@ -499,12 +526,14 @@ actor PlexLiveTVProvider: LiveTVProvider {
             if let ratingKey = tune.ratingKey {
                 items.append(URLQueryItem(name: "rivuletLiveRatingKey", value: ratingKey))
             }
-            start?.queryItems = items
-            if let escapedQuery = start?.percentEncodedQuery?
-                .replacingOccurrences(of: "+", with: "%2B") {
-                start?.percentEncodedQuery = escapedQuery
+            // Carried on the URL rather than threaded through LiveTVDataStore,
+            // matching how the ratingKey above already reaches the keepalive.
+            // `AetherPlayer.needsDeinterlacing` reads it back.
+            if let scan = tune.videoScanType {
+                items.append(URLQueryItem(name: "rivuletLiveScanType", value: scan))
             }
-            return start?.url ?? base
+            start?.queryItems = items
+            return start.flatMap { PlexLiveTVChannel.finalizedLiveURL($0) } ?? base
         } catch {
             SentryBridge.capture(error: error) { scope in
                 scope.setTag(value: "plex_livetv", key: "component")
--- a/Rivulet/Services/Plex/Playback/AetherPlayer.swift
+++ b/Rivulet/Services/Plex/Playback/AetherPlayer.swift
@@ -655,7 +655,12 @@ final class AetherPlayer: PlayerProtocol {
     /// rejects AVPlayer's request pattern).
     func loadLive(url: URL, headers: [String: String]?, forceEngineDemux: Bool = false) async throws {
         let isHLS = Self.liveRoute(for: url, forceEngineDemux: forceEngineDemux) == .nativeHLS
-        let options = LoadOptions(
+        // Forced onto the engine demuxer AND the source is a playlist: the Plex
+        // direct-play case. See the ingest branch below.
+        let usesHLSIngest = !isHLS && Self.isHLSURL(url)
+
+        func makeOptions(nativeRemoteHLS: Bool) -> LoadOptions {
+            LoadOptions(
             suppressDisplayCriteria: false,
             httpHeaders: headers ?? [:],
             // Same rich config as VOD, plus the live-specific flags.
@@ -665,7 +670,7 @@ final class AetherPlayer: PlayerProtocol {
             isLive: true,
             // ~30-minute DVR rewind window (engine retains it disk-backed).
             dvrWindowSeconds: 1800,
-            nativeRemoteHLS: isHLS,
+            nativeRemoteHLS: nativeRemoteHLS,
             preserveASSMarkup: true,
             probesize: 5 * 1024 * 1024,
             maxAnalyzeDuration: 5_000_000,
@@ -678,7 +683,11 @@ final class AetherPlayer: PlayerProtocol {
             // captions on 801 without that flag, so auto-detect returns nothing.
             // Region-default to 801 for AU, otherwise auto-detect.
             teletextPage: Self.regionTeletextPage()
-        )
+            )
+        }
+
+        let options = makeOptions(nativeRemoteHLS: isHLS)
+
         // Same reason as the VOD path: a zap is new content, and broadcast
         // mixes 4:3 SD with 16:9 HD channel to channel, so a reused slot must
         // not measure the new channel against the old one's picture rect.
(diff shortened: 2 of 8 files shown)


l984-451:
> **Created by Claude.** Automated review of `6b08097` against `main`. Engine claims were verified against the pinned AetherEngine 6.5.5 sources, not inferred from the PR description.

## Summary

**Verdict: Request changes.**

The diagnosis behind this PR is correct and valuable. Rivulet really was discarding a granted direct-play session and falling back to AVPlayer, which cannot decode mp2 audio or DVB teletext, and those two are the entire reason to want direct play on a broadcast source. The `%2B` clause-separator bug in `buildStreamURL` is real, and consolidating four build sites behind `finalizedLiveURL` is the right shape for it.

However, the direct-play path as written does not deliver what the description claims. The same-session retry is unreachable, the ingest reader is constructed without client headers, and the multi-tuner guide work is only half wired. Those three are blocking. There is also a scope question at the end.

---

## Blocking

### 1. The `HLSIngestError` retry is dead code

`Rivulet/Services/Plex/Playback/AetherPlayer.swift:736`

The `catch is HLSIngestError` branch never fires, so the encrypted/fMP4 case it exists to handle behaves exactly as it did before, plus a wasted probe.

`HLSLiveIngestReader` does not throw out of `read`. It records the failure into `_terminalError` and returns a negative value (`HLSLiveIngestReader.swift:262`). `avformat_open_input` then fails, and `AetherEngine.load` reaches:

```swift
// AetherEngine.swift:2450
if case .custom = source, !probeOpened {
    state = .error("Failed to load: custom source probe failed")
    throw DemuxerError.openFailed(code: -1)
}
```

That is a `DemuxerError`, not an `HLSIngestError`, so the catch is skipped, the outer catch wraps it as `PlayerError.loadFailed`, and the host fallback ladder runs. That is the second tune this branch was written to avoid, and on a single-tuner box it competes with the grab already held.

Suggested fix: catch the failure broadly on the ingest branch and consult `reader.terminalError` to decide whether to retry natively, rather than matching the error type off `load`.

```swift
let reader = HLSLiveIngestReader(playlistURL: url, httpHeaders: headers ?? [:])
do {
    try await engine.load(source: .custom(reader, formatHint: "mpegts"), options: options)
} catch {
    guard reader.terminalError != nil else { throw error }
    try await engine.load(url: url, startPosition: nil, options: makeOptions(nativeRemoteHLS: true))
}
```

### 2. The ingest reader is constructed without HTTP headers

`Rivulet/Services/Plex/Playback/AetherPlayer.swift:733`

`HLSLiveIngestReader(playlistURL: url)` resolves to the convenience initializer that passes `httpHeaders: [:]` (`HLSLiveIngestReader.swift:107`). A custom `IOReader` performs its own HTTP on its own `URLSession`, so `LoadOptions.httpHeaders` never reaches it. `loadLive` is handed `LiveTVClientIdentity.streamHeaders` and threads them into `LoadOptions`, where they are silently ignored on this path.

AetherEngine's own ingest construction passes them (`HLSLiveIngestReader(playlistURL: candidate, httpHeaders: options.httpHeaders)`), and AE#119 exists specifically for header-enforcing origins. An origin that gates on User-Agent will accept the AVPlayer path and reject this one, surfacing as a generic load failure that looks like an unsupported stream.

```swift
HLSLiveIngestReader(playlistURL: url, httpHeaders: headers ?? [:])
```

### 3. `getLiveTVGuide` still reads only the first DVR

`RivuletCore/Plex/PlexNetworkManager.swift:2576`

`fetchChannelsFromGrid` was changed to iterate every DVR, but the EPG fetch was not:

```swift
guard let dvr = dvrContainer.MediaContainer.Dvr?.first,
      let lineup = dvr.lineup,
      let dvrKey = dvr.key else {
```

A server with two tuners now lists both lineups' channels, but only the first lineup's grid is fetched for programs. Every channel belonging to the second DVR renders with no `PlexLiveTVGuideChannel` match, so its row is empty across the whole timeline. The grouping feature is visible but the guide behind the second tab is blank.

---

## Should fix before merge

### 4. Multiview slots do not get the ingest path

`Rivulet/Views/LiveTV/MultiStreamViewModel.swift:79`

The new routing lives only in `loadLive`. Grid slots call a different method:

```swift
try await aetherPlayer.load(url: url, headers: headers, startTime: nil, isLive: true)
```

That path builds `LoadOptions` with neither `nativeRemoteHLS` nor the ingest branch. Since `resolveStreamURL` is shared, a slot now receives the same `/livetv/sessions/{uuid}/{consumer}/index.m3u8` the fullscreen controller ingests, and the engine's raw-byte live reader detects the `#EXTM3U` body and fails closed with AE#140. The result is that every channel granted direct play works fullscreen and is unplayable in the multiview grid.

### 5. One live-URL rebuild still bypasses `finalizedLiveURL`

`Rivulet/Services/LiveTV/PlexLiveTVProvider.swift:511`

The helper's own documentation states: "Route every build and REBUILD of one of these URLs through here." The direct-play site does not:

```swift
dp.queryItems = items
if let url = dp.url { return url }
```

Reassigning `queryItems` decodes and re-serializes whatever query the PMS-supplied `directPlayPartKey` carried, so any `%2B` in the part key comes back out as a raw `+`. This is exactly the silent failure the helper documents, reintroduced at the one site the sweep missed, and it is on this PR's headline path. Real-world risk is low because part keys rarely carry `+`, but the contract should hold everywhere or it will not hold anywhere.

### 6. Per-DVR grid fetches run sequentially

`RivuletCore/Plex/PlexNetworkManager.swift:2131`

Each loop iteration blocks on `try await request(gridURL, ...)` for a six-hour grid payload before the next DVR is touched, so guide cold-load time scales linearly with tuner count. `GuideLayoutView` shows only a spinner until `channels` is non-empty, so the user sees the full sum. On a remote or relay connection this is seconds per tuner.

A `withThrowingTaskGroup` keyed by DVR index, merged back in index order, preserves the documented first-wins dedupe while overlapping the round trips.

---

## Minor

- **`Rivulet/Services/LiveTV/LiveTVDataStore.swift:789`** — `applyXMLTVChannelLogos` rebuilds `UnifiedChannel` field by field and stops at `isHD`. Because the two new fields have default values, this compiles cleanly while silently resetting `isFavourite` to `false` and `favouriteRank` to `nil` for any channel that picks up an XMLTV logo. Reachability is low, but this is the standard hazard of adding defaulted fields to a struct with hand-rolled copy sites. Worth a `copy`-style helper or an explicit pass-through.

- **`Rivulet/Views/LiveTV/GuideLayoutView.swift:174`** — the category bar is rendered inside `guideContent`, which `channels.isEmpty` replaces with a bare `ProgressView`. `getFavoriteChannelIdentifiers` swallows errors and returns `[]` by design, so a refresh that fails can empty the selected Favourites tab, hide the tab bar along with the grid, and leave the user with a spinner and no way to select another category. The empty-state shape predates this PR, but the PR adds a filter dimension whose contents can vanish independently.

- **`RivuletCore/Models/Plex/PlexLiveTVModels.swift:83`** — `dvrKey` is assigned by the fetch and read nowhere. `resolveStreamURL` still derives the DVR key by splitting the channel's `key` path (`PlexLiveTVProvider.swift:444`). Since this is a new field on a `Codable` model it also joins the synthesized `CodingKeys`. Either route the tune through it or drop it.

- **`RivuletCore/Plex/PlexNetworkManager.swift:3048`** — the new favourite-channel structs were inserted between `PlexLiveTVTuneResult`'s doc comment and its declaration. `/// Result of tuning a Plex Live TV channel.` now attaches to `PlexFavoriteChannelContainer`, and `PlexLiveTVTuneResult` is left undocumented.

- **`firstVideoScanType`** iterates `dict.values`, whose order is not deterministic across runs. It matches the shape of the existing `firstNumericRatingKey`, so it is at least consistent with the file, but if a tune payload ever carries two video streams the winner is arbitrary. Collecting candidates and picking by a stable key would remove the ambiguity.

- **Client profile codec list** — `alac`, `vorbis`, `flac` and `pcm` are declared for an `mpegts` container, which does not carry them, so those entries add no capability. `dts` and `truehd` are meaningful, but note they map to `audioBridgeMode: .lossless`, which means a continuous FLAC bridge encode for the duration of a live session rather than a one-off cost. Worth confirming that trade is intended for a channel that runs for hours.

- **Test coverage** — the added tests cover `guideGroupName` and `lineupName` string cleanup only. The two new pure functions most likely to regress silently, `finalizedLiveURL` and `needsDeinterlacing`, have none. Both are static and take plain values, so they are cheap to cover, and `finalizedLiveURL`'s own doc notes that forgetting the rule is invisible at runtime.

---

## Verified as correct

Two things worth confirming so they do not get re-litigated:

- **Removing the unconditional `setForceSoftwarePathForTesting(true)` is safe.** AetherEngine selects the software backend itself through `VideoRoutingPolicy.requiresSoftwarePath(codecID:fieldOrder:av1Available:spsIndicatesInterlaced:)` (`AetherEngine.swift:2748`), so MPEG-2 sources and streams with a declared interlaced field order still route there regardless of the host flag. The host override only ever removes a force when PMS explicitly reports `progressive`, and absence defaults to interlaced, which preserves current behavior on servers that do not report a scan type.

- **No `#if os(...)` was added to `RivuletCore/`.** The platform boundary holds.

---

## Scope

This PR contains three separable pieces:

1. Direct-play session routing plus the `%2B` encoding fix. A bug fix, clearly in scope.
2. Multi-DVR channel enumeration. Also a bug fix. Taking `.first` was silently dropping a lineup.
3. The Favourites tab. A new feature that introduces a fourth Plex provider host (`epg.provider.plex.tv`) and reaches for the account token inside an actor constructed with a server token.

Items 1 and 2 should merge once the blocking issues are addressed. Item 3 is small, self-contained, and maps to a real Plex concept the user arranged by hand, so it is defensible, but it would be better as a separate PR. As bundled, a problem with the new provider host cannot be reverted without also losing the playback fix.

PR #311, opened 2026-09-03
Live TV: Support pipe-delimited stream URL headers and remove iOS UA setting

tresby:
### Summary
1. **Support Pipe-Delimited Stream Headers (`|User-Agent=...`)**:
   * Supports the standard IPTV syntax used by players like TiviMate and Kodi (`<URL>|User-Agent=<UA>&Referer=<Referer>`).
   * `M3UParser` extracts pipe-delimited headers into `httpHeaders` and sanitizes the stream URL before initializing `ParsedChannel`.
   * Both tvOS and iOS players (`AetherPlayer.loadLive`, `MultiStreamViewModel`, and `IOSAetherPlayerView`) forward custom headers and sanitize any runtime stream URLs via `LiveTVClientIdentity.resolveStream(url:baseHeaders:)`.

2. **Remove User-Agent Setting from iOS Tab**:
   * Removed the explicit User-Agent text input from the iOS Live TV Source Settings screen.
   * Streams identify with the standard `LiveTVClientIdentity.userAgent` (`Rivulet/<version>`) unless a custom User-Agent is authored directly in the stream link via `|User-Agent=`.

### Testing
* Added unit tests in `M3UParserTests` and `DispatcharrURLBuilderTests` for URL sanitization and header extraction.
* Built and passed all unit tests on tvOS and iOS simulators.

### DIFF
--- a/Rivulet/Services/Plex/Playback/AetherPlayer.swift
+++ b/Rivulet/Services/Plex/Playback/AetherPlayer.swift
@@ -703,10 +703,13 @@ final class AetherPlayer: PlayerProtocol {
     /// native path fails against a server (e.g. a Plex transcode session that
     /// rejects AVPlayer's request pattern).
     func loadLive(url: URL, headers: [String: String]?, forceEngineDemux: Bool = false) async throws {
-        let isHLS = Self.liveRoute(for: url, forceEngineDemux: forceEngineDemux) == .nativeHLS
+        let resolved = LiveTVClientIdentity.resolveStream(url: url, baseHeaders: headers ?? [:])
+        let streamURL = resolved.url
+        let streamHeaders = resolved.headers
+        let isHLS = Self.liveRoute(for: streamURL, forceEngineDemux: forceEngineDemux) == .nativeHLS
         let options = LoadOptions(
             suppressDisplayCriteria: false,
-            httpHeaders: headers ?? [:],
+            httpHeaders: streamHeaders,
             // Same rich config as VOD, plus the live-specific flags.
             matchContentEnabled: true,
             panelIsInHDRMode: Self.panelIsInHDRMode(),
--- a/Rivulet/Views/LiveTV/LiveTVAetherPlayerViewController.swift
+++ b/Rivulet/Views/LiveTV/LiveTVAetherPlayerViewController.swift
@@ -322,10 +322,16 @@ final class LiveTVAetherPlayerViewController: UIViewController {
             isNativeHLSRoute = route == .nativeHLS
             joinTelemetry?.resolveFinished(url: url, route: route)
             startLiveSessionKeepAlive(for: url)
+            var headers = LiveTVClientIdentity.streamHeaders
+            if let customHeaders = channel.httpHeaders {
+                for (k, v) in customHeaders {
+                    headers[k] = v
+                }
+            }
             do {
                 try await aether.loadLive(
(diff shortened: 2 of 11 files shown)


l984-451:
### Code review

The parse-time half is right: a `url|User-Agent=...` line currently reaches the engine with the pipe percent-encoded into the path, so those playlists 404. Parsing at the M3U seam is the correct place, and putting `LiveTVClientIdentity` in `RivuletCore` is the right home for it. Three blocking items, all in the runtime layer around it.

**1. `loadLive` sanitizes the URL and then loads the unsanitized one.** `streamURL` is used to pick the HLS route, but `engine.load` gets the original `url`. When `resolveStream` does anything, the route is decided from `.../ch1.m3u8` while the engine is handed `.../ch1.m3u8%7CUser-Agent=...`. Same shape in the iOS copy at `RivuletiOS/Player/AetherPlayer.swift#L262-L290`. The test covers the pure function, so it passes while the caller discards the result.

https://github.com/l984-451/Rivulet/blob/d491c27098b6becc4d2d472e66a5f4328ee94c05/Rivulet/Services/Plex/Playback/AetherPlayer.swift#L705-L708

**2. The XMLTV logo pass drops the parsed headers.** `applyXMLTVChannelLogos` rebuilds `UnifiedChannel` field by field and omits `httpHeaders`, so it defaults to nil. A channel with no `tvg-logo` loads with its header, then the EPG pass replaces it with a headerless copy and every tune after that sends the default UA. It compiles silently because the init parameter is defaulted.

https://github.com/l984-451/Rivulet/blob/d491c27098b6becc4d2d472e66a5f4328ee94c05/Rivulet/Services/LiveTV/LiveTVDataStore.swift#L788-L802

**3. The `%7C` fallback truncates legitimate URLs.** Kodi and TiviMate split on the literal pipe only; an encoded `%7C` is ordinary URL data. `http://cdn/live.m3u8?token=abc%7Cdef&expires=170...` becomes `http://cdn/live.m3u8?token=abc` plus a bogus `expires` header, so a signed URL that works today starts returning 403. Same for an Xtream password containing a pipe.

https://github.com/l984-451/Rivulet/blob/d491c27098b6becc4d2d472e66a5f4328ee94c05/RivuletCore/LiveTV/LiveTVClientIdentity.swift#L52-L56

Deleting the runtime `resolveStream` layer resolves 1 and 3 together: every Live TV URL reaching `loadLive` has already been through `M3UParser`, and Plex live URLs contain no pipes. One stripper at the parse seam, no drift.

Smaller things while you are in here:

- `parseStreamURL` is called from the nonisolated XML parser and both build jobs now warn `call to main actor-isolated static method ... in a synchronous actor-isolated context`. It is pure, so mark it `nonisolated`.
- iOS keeps only User-Agent, Referer and Authorization, so `|Origin=...` or `|Cookie=...` plays on tvOS and 403s on iOS from the same playlist.
- The base-header merge is hand-rolled at five call sites, against the doc comment on `streamHeaders` that says it is one value so the call sites cannot drift apart.

Please split the iOS User-Agent setting removal into its own PR. `playbackHeaders` is used for the playlist and XMLTV fetches too, not just playback, so a provider that gates on UA can no longer be reached at all, and the pipe syntax cannot rescue that because it lives inside the playlist you can no longer fetch.

---
Reviewed by Claude Code at the maintainer's request.

PR #313, opened 2026-09-08
Live TV: smart EPG image resolution and role-based fallback for tvOS and iOS

tresby:
# Live TV: smart EPG image resolution and role-based fallback for tvOS and iOS

### Summary
XMLTV feeds often provide only a single landscape image for a programme (sometimes declared with `width`/`height` attributes, sometimes unlabelled). Previously, these landscape images were placed directly into the 2:3 portrait poster slot in the guide header, distorting or pillarboxing the artwork while leaving the guide backdrop empty or falling back inappropriately.

This PR introduces aspect-ratio classification and platform-tailored artwork routing:
- **tvOS**: Landscape images (aspect ratio ≥ 1.25) route exclusively to the guide background backdrop. The 2:3 poster slot never renders landscape images; if no portrait artwork exists, it falls back to the **channel logo** (`channel.logoURL`). When no landscape artwork exists, the backdrop falls back to the canonical neutral charcoal vignette settings gradient (`GuideDefaultBackdrop`).
- **iOS**: Programme detail prioritises 16:9 landscape (`landscapeURL`), falls back to 2:3 portrait poster (`posterURL`), and **never falls back to the channel logo**. The background uses the same `GuideDefaultBackdrop` gradient.
- **On-the-fly Inspection & Performance**: Dynamically inspects unlabelled images on load via existing shared in-memory caches (`ImageCacheManager` / `IOSArtworkCache`). Fully decoupled from global image decoding, with zero disk I/O, zero detached tasks, and zero duplicate network fetches.

### Key Changes
- **`EPGImageClassifier` (`RivuletCore/LiveTV/EPGImageClassifier.swift`)**:
  - Lightweight, thread-safe in-memory cache (`NSLock`) tracking aspect ratio classifications (`.landscape` vs `.portrait`).
  - Purely synchronous lookup; no disk/`UserDefaults` persistence or raw network requests.
- **`XMLTVParser` (`RivuletCore/IPTV/XMLTVParser.swift`)**:
  - `posterIcon(from:)`: Selects icons with aspect ratio < 1.25 (or pre-classified portrait). Returns `nil` if only landscape images are provided so the UI can fall back to the channel logo (tvOS) or landscape backdrop (iOS).
  - `landscapeIcon(from:)`: Selects icons with aspect ratio ≥ 1.25 (or pre-classified landscape).
- **tvOS Guide (`EPGGuideView.swift` & `GuideLayoutView.swift`)**:
  - `GuideInfoBar`: The 2:3 poster box strictly renders portrait artwork. If an image is landscape (or resolves as landscape upon loading), it immediately swaps to the channel logo (`logoInset`).
  - `GuideLayoutView`: Backdrops load landscape artwork when available. If no landscape image exists or is still resolving, `GuideDefaultBackdrop` displays the app's settings gradient.
- **iOS Programme Detail (`IOSGuideView.swift`)**:
  - `IOSProgramDetailView`: Prioritises `landscapeURL` → pre-classified landscape → resolved landscape → `posterURL` → `iconURL`. Never falls back to the channel logo.
  - `IOSProgramDetailArtworkBackdrop`: Uses `GuideDefaultBackdrop` behind the image and gradient scrims.
- **Shared App Gradient (`GuideDefaultBackdrop.swift`)**:
  - Extracted the neutral charcoal vignette gradient from the app settings menu into a reusable component for tvOS and iOS.

### Verification
- **Unit Tests**: `XMLTVParserTests` (tvOS simulator) — 26 passed, 0 failed.
- **Builds**: Cleanly compiled both `Rivulet` (tvOS) and `Rivulet iOS`.


l984-451:
### Code review

The goal is right: a 16:9 still squeezed into a 168x252 slot is worse than the channel logo, and that art belongs on the backdrop. The mechanism is where it goes wrong.

**1. The classifier is main-actor isolated and called from the nonisolated parser.** Both app targets set `SWIFT_DEFAULT_ACTOR_ISOLATION = MainActor`, so a class with no explicit isolation is `@MainActor`, and `@unchecked Sendable` does not opt out of that. `XMLTVInternalParser` is `nonisolated`, so the calls at `XMLTVParser.swift#L180`, `#L277` and `#L296` warn today and are errors under language mode 6. One word fixes it: `nonisolated final class`.

https://github.com/l984-451/Rivulet/blob/1be10382ab361ae3485848937a7fc9113e0891c1/RivuletCore/LiveTV/EPGImageClassifier.swift#L13-L16

**2. The parser's output now depends on global mutable UI state.** `posterIcon` and `landscapeIcon` consult a process-global registry that the views write to, so the same XMLTV bytes parse differently depending on what the user browsed. Focus a programme, the view downloads and registers its icon as landscape, the guide refreshes the EPG, and the info-bar poster silently changes to the channel logo with no user action. The registry is also never evicted or cleared, holding a `URL` per unique icon for the life of the process. Within a single parse the lookup adds almost nothing, since declared `width`/`height` already handled that case two lines above. Dropping the classifier lookups out of the parser keeps the good half of that diff.

**3. Channel logos are now centre-cropped.** The post-load check applies `.scaleAspectFill` to any image with a ratio at or above 1.25, including `channel.logoURL`, which is what gets picked whenever no landscape art is known. A 1000x400 wordmark logo gets its sides cropped and the channel becomes unidentifiable, which is what the doc comment directly above forbids: "The channel logo is preferred over a poster as the fallback: in a channel list, identifying the channel beats showing a sliver of programme art." The guard wants to be `candidateLandscape != nil`. The comment at `#L357-L365` is also stale now, it still says 1.3.

https://github.com/l984-451/Rivulet/blob/1be10382ab361ae3485848937a7fc9113e0891c1/Rivulet/Views/LiveTV/UIKit/ChannelListPanelView.swift#L376-L380

**4. The guide can show the previous programme's backdrop.** `resolvedLandscapeURL` is a bare `@State` not keyed to a programme, and `guideBackdropURL` falls through to it, so `.task(id:)` does not re-fire when focus moves to an unclassified programme. The previous artwork stays behind the new info bar for the whole classify download, which on a dead host is three attempts with backoff. Before this it went to nil and faded within the 180ms settle. Storing `(programID, URL)` and ignoring a resolution whose id no longer matches fixes it. Same family as `a65b6b8`.

https://github.com/l984-451/Rivulet/blob/1be10382ab361ae3485848937a7fc9113e0891c1/Rivulet/Views/LiveTV/GuideLayoutView.swift#L239-L241

**5. Please split `GuideDefaultBackdrop` into its own PR.** On tvOS this replaces a real system background with a hand-rolled gradient, and the comment at `GuideLayoutView.swift#L137-L139` still says the stock background shows through. The gradient is the iOS mimic of the tvOS surface, so tvOS ends up approximating itself. It is also the first SwiftUI `View` in `RivuletCore`. That is a look change rather than an image-routing one, and it wants a device screenshot.

Good call putting the classifier in `RivuletCore`. The part that stayed behind is the download-measure-register block, which is pasted verbatim three times (`EPGGuideView.swift#L1543-L1560`, `GuideLayoutView.swift#L259-L279`, `IOSGuideView.swift#L171-L186`), differing only in which cache it calls. That one wants to be `classify(url:) async` on `EPGImageClassifier` with an injected loader. The role selection can stay per-platform, since a tvOS 2:3 poster slot and an iOS full-screen backdrop genuinely want different answers, but the 1.25 threshold appears at eight sites in six files and should be one constant next to the classifier.

On iOS, `IOSGuideView` drops the orientation-aware choice, so a portrait phone now centre-crops a 16:9 image to about 26% of its width where a 2:3 poster used to fill the frame. For Plex Live TV both are populated, so every portrait user loses the poster. Intentional, but worth saying so in the description.

---
Reviewed by Claude Code at the maintainer's request.

